PR #48, opened 2021-04-10
set dart version to stable

sitaram-kalluri:


sitaram-kalluri:
The changes are pushed to trunk branch from a different PR.

PR #125, opened 2021-12-02
update README.md in at_lookup

sitaram-kalluri:
**- What I did** Update the README.md content for at_lookup package.

**- How I did it**

**- How to verify it**


at_lookup/README.md
```diff
@@ -3,48 +3,108 @@
...
 
 # at_lookup library
-The AtLookup Library is the low-level direct implementation of the @protocol verbs
 
-## Installation:
-To use this library in your app, first add it to your pubspec.yaml
-```  
+## Overview:
+
+The AtLookup Library is the low-level direct implementation of the @protocol verbs. The AtClient packages depends on
```
murali-shris: we need not mention atclient here..similar to at_client there may be other packages dependent on this

at_lookup/README.md
```diff
@@ -3,48 +3,108 @@
...
+this package to execute commands(scan, update, lookup, llookup, plookup, etc) on the secondary server.
+
+## Get started:
+
+### Installation:
+
+To use this library in your app, add it to your pubspec.yaml
```
murali-shris: add "the dependency"

gkc (commented):
How often would a developer need to use this library directly? Should we add a comment saying something like, "In most cases while developing client code that uses the @ protocol, you will use the at_client or at_client_flutter packages. This package provides direct access to the lower-level @ protocol implementation should you need it."

@nickelskevin wdyt?

cpswan (approved):

at_lookup/example/README.md
```diff
@@ -0,0 +1,13 @@
+<img src="https://atsign.dev/assets/img/@developersmall.png?sanitize=true">
```
cpswan: A lot of what's here should be back in the README.md because it's about the package rather than the example.

Take a look at [archetype-pubdev](https://github.com/atsign-foundation/archetype-pubdev) for some examples. Also we're switching to the new @ platform logo.

sitaram-kalluri: Updated the README.md content and the @platform logo

at_lookup/example/README.md
```diff
@@ -0,0 +1,11 @@
+<img src="https://atsign.dev/assets/img/@platform_logo_grey.svg?sanitize=true">
+
+### Now for a little internet optimism
+
+# at_lookup_example library
+
+The atlookup package is an interface between the Client SDK and the cloud secondary.
+
```
cpswan, line 8: It would be good to have the code snippet that was previously in the README (and main.dart) in this file.

cpswan (changes requested):

at_lookup/example/README.md
sitaram-kalluri: Added code snippet.

cpswan (approved):

PR #136, opened 2022-02-15
Implement error handling

sitaram-kalluri:
**- What I did** 

1. Handle `error:` responses from secondary server and throw AtLookupException.
2. Throw `AtTimeoutException` when server fails to respond back or for network issues.
3. Convert the internal methods to null safe.

### DIFF
--- a/at_lookup/lib/src/at_lookup_impl.dart
+++ b/at_lookup/lib/src/at_lookup_impl.dart
@@ -328,8 +328,18 @@ class AtLookupImpl implements AtLookUp {
     // If response starts with error:, throw AtLookupException.
     if (_isError(verbResult)) {
       verbResult = verbResult.replaceAll('error:', '');
-      throw AtLookUpException(
-          verbResult.split('-')[0], verbResult.split('-')[1]);
+      // Setting the errorCode and errorDescription to default values.
+      var errorCode = 'AT0014';
+      var errorDescription = 'Unknown server error';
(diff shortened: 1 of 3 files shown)


gkc (approved):
LGTM

at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -319,25 +317,31 @@ class AtLookupImpl implements AtLookUp {
...
+      throw AtLookUpException('AT0014', 'Request timed out');
+    }
+    // If response starts with error:, throw AtLookupException.
     if (_isError(verbResult)) {
       verbResult = verbResult.replaceAll('error:', '');
-      var errorCode = verbResult.split('-')[0];
-      var errorMessage = verbResult.split('-')[1];
-      return Future.error(AtLookUpException(errorCode, errorMessage));
+      throw AtLookUpException(
+          verbResult.split('-')[0], verbResult.split('-')[1]);
```
gkc: Please add defensive code for when the verbResult doesn't contain '-' just in case in future we have a badly behaved server which returns an error code without an error description

gkc (changes requested):

at_lookup/lib/src/at_lookup_impl.dart
sitaram-kalluri: Added a defensive code when server does not return error response with '-'

murali-shris (approved):

PR #150, opened 2022-03-23
Github workflows for at_onboarding_cli

srieteja:
**- What I did**
adding git workflows to test at_onboarding_cli

**- How I did it**

**- How to verify it**


srieteja:
@cpswan both the 'check_docker_readiness' and 'check_test_env' seem to be failing. They are retrying a couple of times and then timing out after 5mins.

This is the response from ping:

Run ping -c 10 vip.ve.atsign.zone -p 64
  ping -c 10 vip.ve.atsign.zone -p 64
  shell: /usr/bin/bash -e {0}
  env:
    PUB_CACHE: /home/runner/.pub-cache
    DART_HOME: /opt/hostedtoolcache/dart-sdk
PATTERN: 0x64
PING vip.ve.atsign.zone (10.64.64.64) 56(84) bytes of data.

--- vip.ve.atsign.zone ping statistics ---
10 packets transmitted, 0 received, 100% packet loss, time 9[2](https://github.com/atsign-foundation/at_libraries/runs/5681993964?check_suite_focus=true#step:15:2)0[3](https://github.com/atsign-foundation/at_libraries/runs/5681993964?check_suite_focus=true#step:15:3)ms

cpswan:
@srieteja I'm looking at this now.

One of the tests used by at_client_sdk depends on running a script against the virtual environment to load data. So I'd expect that to fail.

Do you have things working locally?

.github/workflows/at_libraries.yaml
```diff
@@ -53,3 +53,49 @@ jobs:
...
+        run: dart pub get
+      
+      # Run dart analyzer in at_cli_onboarding
+      - name: Run dart analyzer in at_onboarding_cli
+        working-directory: at_onboarding_cli
+        run: dart analyze
+      
+      #functional tests for at_onbaording_cli
+      - name: add entry to hosts file
+        run: echo echo "127.0.0.1    vip.ve.atsign.zone" | sudo tee -a /etc/hosts
```
cpswan: `echo echo` might be causing problems here

.github/workflows/at_libraries.yaml
```diff
@@ -53,3 +53,49 @@ jobs:
...
+        working-directory: at_onboarding_cli/functional_tests
+        run: dart pub get
+        
+      - name: Start docker instance
+        working-directory: at_onboarding_cli/functional_tests
+        run: sudo docker-compose up -d
+        
+#       - name: Check for docker container readiness
+#         working-directory: at_onboarding_cli/functional_tests
+#         run: dart run check_docker_readiness.dart
```
cpswan: This works for me locally

.github/workflows/at_libraries.yaml
```diff
@@ -53,3 +53,49 @@ jobs:
...
+
+      - name: ping
+        run: ping -c 10 vip.ve.atsign.zone -p 64
+        
+      - name: run pkamLoad on docker-image
```
cpswan, line 92: Is this needed for these tests?

.github/workflows/at_libraries.yaml
```diff
@@ -53,3 +53,49 @@ jobs:
...
+        run: sudo docker exec functional_tests_virtualenv_1 supervisorctl start pkamLoad
+      
+#       - name: Check test environment readiness
+#         working-directory: at_onboarding_cli/functional_tests
+#         run: dart run check_test_env.dart
```
cpswan: This does need the PKAM load to succeed (and won't work if the name resolution is messed up).

cpswan (changes requested):
I think the `echo echo` is causing a bad entry to /etc/hosts, which then prevents vip.ve.atsign.zone name resolution from working as it should.

cpswan:
The logs confirm that vip.ve.atsign.zone is being resolved by DNS to 10.64.64.64 rather than hosts to 127.0.0.1:

```
Run ping -c 10 vip.ve.atsign.zone -p 64
  ping -c 10 vip.ve.atsign.zone -p 64
  shell: /usr/bin/bash -e {0}
  env:
    PUB_CACHE: /home/runner/.pub-cache
    DART_HOME: /opt/hostedtoolcache/dart-sdk
PATTERN: 0x64
PING vip.ve.atsign.zone (10.64.64.64) 56(84) bytes of data.

--- vip.ve.atsign.zone ping statistics ---
10 packets transmitted, 0 received, 100% packet loss, time 9[2](https://github.com/atsign-foundation/at_libraries/runs/5682044466?check_suite_focus=true#step:15:2)[10](https://github.com/atsign-foundation/at_libraries/runs/5682044466?check_suite_focus=true#step:15:10)ms
```

.github/workflows/at_libraries.yaml
srieteja: Ahh my bad, removing this solved everything.

cpswan (approved):

PR #169, opened 2022-05-12
Parse server error responses

sitaram-kalluri:
**- What I did**
Parse JSON server responses in at_lookup

### DIFF
--- a/at_lookup/lib/src/at_lookup_impl.dart
+++ b/at_lookup/lib/src/at_lookup_impl.dart
@@ -267,6 +267,19 @@ class AtLookupImpl implements AtLookUp {
     }
     // If response starts with error:, throw AtLookupException.
     if (_isError(verbResult)) {
+      // If server returns the error response in JSON Encoded string, decodes the JSON String
+      // and throws the appropriate AtException basing on the error code.
+      if (verbResult.contains('errorCode')) {
+        var decodedResponse = jsonDecode(verbResult);
+        throw AtExceptionUtils.get(
+            decodedResponse['errorCode'], decodedResponse['errorDescription'])
+          ..stack(Triple(
+              Intent.remoteVerbExecution,
+              ExceptionScenario.remoteVerbExecutionFailed,
+              decodedResponse['errorDescription']));
+      }
+      // If server returns the error response in String.
+      // Preserves the backward compatibility
       verbResult = verbResult.replaceAll('error:', '');
       // Setting the errorCode and errorDescription to default values.
       var errorCode = 'AT0014';
@@ -286,7 +299,7 @@ class AtLookupImpl implements AtLookUp {
   }
 
   bool _isError(String verbResult) {
-    return verbResult.startsWith('error:');
+    return verbResult.startsWith('error:') || verbResult.contains('errorCode');
   }
 
   Future<String> _update(UpdateVerbBuilder builder) async {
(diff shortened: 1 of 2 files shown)


at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -267,6 +267,19 @@ class AtLookupImpl implements AtLookUp {
     }
     // If response starts with error:, throw AtLookupException.
     if (_isError(verbResult)) {
+      // If server returns the error response in JSON Encoded string, decodes the JSON String
+      // and throws the appropriate AtException basing on the error code.
+      if (verbResult.contains('errorCode')) {
```
murali-shris: contains check may fail if some random data from server has errorCode in it..
error: < text > - old format
error: < json > - new format
_isError(verbResult)) can have the previous implementation..if json parsing fails, then consider error response to be in old format

VJag (approved):

at_lookup/lib/src/at_lookup_impl.dart
sitaram-kalluri: Modified the code to have error: before JSON response.

at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -267,19 +267,31 @@ class AtLookupImpl implements AtLookUp {
...
+        var decodedResponse = jsonDecode(verbResult);
+        throw AtExceptionUtils.get(
+            decodedResponse['errorCode'], decodedResponse['errorDescription'])
+          ..stack(Triple(
+              Intent.remoteVerbExecution,
+              ExceptionScenario.remoteVerbExecutionFailed,
+              decodedResponse['errorDescription']));
+        // If JSON decoding fails, fall back to older version
+        // Parse the error response in throw AtLookupException.
+      } on FormatException {
```
gkc: I don't know how expensive it is to try to decode and then throw a FormatException. Could perhaps reduce the likelihood of that happening by checking first of all if the verbResult starts with a '{'? (and only trying to jsonDecode if it does indeed look like json)

gkc (changes requested):

at_lookup/lib/src/at_lookup_impl.dart
sitaram-kalluri: Added IF condition to verify if result startsWith '{' and endsWith '}' to check if response is JSON encoded string.

at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -267,7 +267,20 @@ class AtLookupImpl implements AtLookUp {
...
     // If response starts with error:, throw AtLookupException.
     if (_isError(verbResult)) {
-      verbResult = verbResult.replaceAll('error:', '');
+      verbResult = verbResult.replaceFirst('error:', '');
+      // If condition to verify if verbResult is a JSON Encoded String.
+      if (verbResult.startsWith('{') && verbResult.endsWith('}')) {
+        try {
+          var decodedResponse = jsonDecode(verbResult);
+          throw AtExceptionUtils.get(decodedResponse['errorCode'],
+              decodedResponse['errorDescription']);
```
gkc, line 276: This method (get(errorCode, errorDescription) isn't in the current AtExceptionUtils in at_tools ... can you tell me which PR it is in?

sitaram-kalluri: It is in at_commons. Following is the PR: https://github.com/atsign-foundation/at_tools/pull/186/files

gkc (approved):

sitaram-kalluri:
Parsing the error responses are deferred to next releases. Hence closing the PR.

PR #182, opened 2022-06-21
feat: Add network checker before writing to socket

sitaram-kalluri:
**- What I did**
1. Add check network connectivity before writing command to server. If network connectivity fails throws `AtConnectException`

Fixes the [issue-546](https://github.com/atsign-foundation/at_client_sdk/issues/546), where at_lookup throws exception when there is no network and the apps might have to initiate a retry.

**- How I did it**

**- How to verify it**


sitaram-kalluri:
The changes are not valid. Hence closing the PR.

PR #185, opened 2022-06-28
fix: handle partial prompt response and add unit tests

sitaram-kalluri:
With the current logic in `messageHandler`, if a partial prompt response is received, then the partial responses gets appended to subsequent responses.

**- What I did**
1. Refactor the messageHandler code as follows:
    - Add data to the buffer until we encounter a terminating character (new line character - \n)
    - Decode the data to the String format.
    - Trim the prompt the from the String format
    - Check if response as partial prompt appended from the previous responses. If yes trim it as well

### DIFF
--- a/at_lookup/lib/src/connection/outbound_message_listener.dart
+++ b/at_lookup/lib/src/connection/outbound_message_listener.dart
@@ -8,48 +8,88 @@ import 'package:at_utils/at_logger.dart';
 ///Listener class for messages received by [RemoteSecondary]
 class OutboundMessageListener {
   final logger = AtSignLogger('OutboundMessageListener');
-  final _buffer = ByteBuffer(capacity: 10240000);
-  late Queue _queue;
+  late ByteBuffer _buffer;
+  final Queue _queue = Queue();
   final _connection;
   Function? syncCallback;
+  final int newLineCodeUnit = 10;
+  final int atCharCodeUnit = 64;
 
-  OutboundMessageListener(this._connection);
+  OutboundMessageListener(this._connection, {int bufferCapacity = 10240000}) {
+    _buffer = ByteBuffer(capacity: bufferCapacity);
+  }
 
   /// Listens to the underlying connection's socket if the connection is created.
   /// @throws [AtConnectException] if the connection is not yet created
   void listen() {
-    _connection.getSocket().listen(_messageHandler,
+    _connection.getSocket().listen(messageHandler,
         onDone: _finishedHandler, onError: _errorHandler);
-    _queue = Queue();
   }
 
   /// Handles messages on the inbound client's connection and calls the verb executor
   /// Closes the inbound connection in case of any error.
   /// Throw a [BufferOverFlowException] if buffer is unable to hold incoming data
-  Future<void> _messageHandler(data) async {
+  Future<void> messageHandler(List data) async {
     String result;
-    if (!_buffer.isOverFlow(data)) {
-      // skip @ prompt
-      if (data.length == 1 && data.first == 64) {
-        return;
-      }
-      //ignore prompt(@ or @<atSign>@) after '\n'
-      // If data contains 10(utf8 character for '\n'), trim the string after the '\n'
-      if (data.contains(10)) {
-        data = data.sublist(0, data.lastIndexOf(10) + 1);
-      }
-      _buffer.append(data);
+    var offset;
+    // check buffer overflow
+    _checkBufferOverFlow(data);
+
+    // If the data contains a new line character, add until the new line char to buffer
+    if (data.contains(newLineCodeUnit)) {
+      offset = data.lastIndexOf(newLineCodeUnit);
+      var dataSubList = data.getRange(0, offset).toList();
+      _buffer.append(dataSubList);
     } else {
+      offset = 0;
+    }
+    // Loop from last index to until the end of data.
+    // If a new line character and followed by @ character is found, then it is end
+    // of server response. process the data.
+    // Else add the byte to buffer.
+    for (int element = offset; element < data.length; element++) {
+      // If element is @ character and lastCharacter in the buffer is \n,
+      // then complete data is received. process it.
+      if (data[element] == atCharCodeUnit &&
+          _buffer.getData().last == newLineCodeUnit) {
+        // remove the terminating character (last \n) from the server response.
+        // preserve other new line characters.
+        List<int> temp = (_buffer.getData().toList())..removeLast();
+        result = utf8.decode(temp);
+        result = _stripPrompt(result);
+        _queue.add(result);
+        //clear the buffer after adding result to queue
+        _buffer.clear();
+        _buffer.addByte(data[element]);
+      } else {
+        _buffer.addByte(data[element]);
+      }
+    }
+  }
+
+  /// The methods verifies if buffer has the capacity to accept the data.
+  ///
+  /// Throw BufferOverFlowException if data length exceeds the buffer capacity
+  _checkBufferOverFlow(data) {
+    if (_buffer.isOverFlow(data)) {
+      int bufferLength = (_buffer.length() + data.length) as int;
       _buffer.clear();
       throw BufferOverFlowException(
-          'Buffer overflow on outbound connection result');
+          'data length exceeded the buffer limit. Data length : $bufferLength and Buffer capacity ${_buffer.capacity}');
     }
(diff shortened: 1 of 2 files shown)


at_lookup/lib/src/connection/outbound_message_listener.dart
```diff
@@ -27,34 +31,58 @@ class OutboundMessageListener {
...
-        data = data.sublist(0, data.lastIndexOf(10) + 1);
-      }
-      _buffer.append(data);
-    } else {
+    // Throw exception if data length exceeds the buffer limit.
+    if (_buffer.isOverFlow(data)) {
+      int bufferLength;
+      (_buffer.length() == 0)
+          ? bufferLength = data.length
+          : bufferLength = (_buffer.length() + data.length) as int;
```
gkc: I think the above can be simplified to `bufferLength = (_buffer.length() + data.length) as int;`

at_lookup/lib/src/connection/outbound_message_listener.dart
```diff
@@ -27,34 +31,58 @@ class OutboundMessageListener {
...
+
+  /// The method accepts the result (server response) and trim's the prompt from the response
+  /// and returns the actual response.
+  String _trimPrompt(String result) {
+    var newLineIndex = result.lastIndexOf('\n');
+    var trimmedResult = result.substring(0, newLineIndex);
+    var prompt = result.substring(newLineIndex).trim();
+    // If isInCompletePromptReceived is true, then partial prompt from previous response
+    // is added to data. trim the prompt from result and reset _isInCompletePromptReceived to false.
+    if (_isInCompletePromptReceived) {
```
gkc: Should never be possible. The whole point of waiting for newline is that once we receive a full response line terminated by '\n', we *know* that there is a prompt at the start of the line, which we need to trim.

We have two types of info available to us to help us easily trim the prompt

1) We know that messages from the server, after the prompt portion, always start with `data:` or `error:` or `notification:`. The Java SDK uses this information
2) We know that if we have an unauthenticated connection the prompt will be '@' but if we have an authenticated connection the prompt will be '@<atSign>@'

at_lookup/lib/src/connection/outbound_message_listener.dart
```diff
@@ -27,34 +31,58 @@ class OutboundMessageListener {
...
+    result = utf8.decode(_buffer.getData());
+    result = _trimPrompt(result);
+    _queue.add(result);
+    //clear the buffer after adding result to queue
+    _buffer.clear();
+  }
+
+  /// The method accepts the result (server response) and trim's the prompt from the response
+  /// and returns the actual response.
+  String _trimPrompt(String result) {
```
gkc: Implementation is more complex than it needs to be. See comment below

gkc (changes requested):

at_lookup/lib/src/connection/outbound_message_listener.dart
sitaram-kalluri: Refactored the code and used 'data:' or 'error:' or 'stream:' to find the start of server responses.

at_lookup/lib/src/connection/outbound_message_listener.dart
sitaram-kalluri: Refactored the code as mentioned.

at_lookup/lib/src/connection/outbound_message_listener.dart
sitaram-kalluri: Refactored the code and pushed the changes as mentioned.

at_lookup/lib/src/connection/outbound_message_listener.dart
```diff
@@ -8,50 +8,76 @@ import 'package:at_utils/at_logger.dart';
...
-      // If data contains 10(utf8 character for '\n'), trim the string after the '\n'
-      if (data.contains(10)) {
-        data = data.sublist(0, data.lastIndexOf(10) + 1);
-      }
-      _buffer.append(data);
-    } else {
+    // check buffer overflow
+    _checkBufferOverFlow(data);
+    // Add data to buffer until a new line character is encountered
+    _buffer.append(data);
```
gkc: You need to process the data one byte at a time in order to avoid bug with situation where there is more than one newLine

at_lookup/lib/src/connection/outbound_message_listener.dart
```diff
@@ -8,50 +8,76 @@ import 'package:at_utils/at_logger.dart';
...
+    // If data does contains terminating character process the data in buffer
+    if (data.contains(terminatingChar)) {
+      result = utf8.decode(_buffer.getData());
```
gkc: Nope, as I said before, you can only decode what you know is decodable - i.e. everything up to the newline

See the algorithm I proposed in an earlier review

gkc (changes requested):

at_lookup/lib/src/connection/outbound_message_listener.dart
```diff
@@ -8,48 +8,76 @@ import 'package:at_utils/at_logger.dart';
...
-      result = result.trim();
-      _buffer.clear();
-      _queue.add(result);
+  }
+
+  /// The method accepts the result (server response) and trim's the prompt from the response
+  /// and returns the actual response.
+  String _stripPrompt(String result) {
+    // trim the \n character in the last.
+    result = result.trim();
```
gkc: Please only trim the \n

at_lookup/lib/src/connection/outbound_message_listener.dart
```diff
@@ -8,48 +8,76 @@ import 'package:at_utils/at_logger.dart';
...
+    for (var element in data) {
+      // If element is @ character and lastCharacter in the buffer is \n,
+      // then complete data is received. process it.
+      if (element == atCharCodeUnit &&
+          _buffer.getData().last == terminatingChar) {
+        result = utf8.decode(_buffer.getData());
+        result = _stripPrompt(result);
+        _queue.add(result);
+        //clear the buffer after adding result to queue
+        _buffer.clear();
```
gkc, line 62: Suggest we now add `_buffer.addByte(element);` here as well, so that the '@' is at the start of the next buffer

at_lookup/lib/src/connection/outbound_message_listener.dart
```diff
@@ -8,48 +8,76 @@ import 'package:at_utils/at_logger.dart';
...
-      //ignore prompt(@ or @<atSign>@) after '\n'
-      // If data contains 10(utf8 character for '\n'), trim the string after the '\n'
-      if (data.contains(10)) {
-        data = data.sublist(0, data.lastIndexOf(10) + 1);
+    // check buffer overflow
+    _checkBufferOverFlow(data);
+    //1. Iterate the list until a new line character is encounter.
+    //2. If new line char is encountered, then decode the buffered data (utf8.decode)
+    //3. Strip the prompt from the result.
+    //4. Add the result to queue.
```
gkc: We used to look at the whole data and break it on '\n'

What if we now do this:
```
Look at the whole data and look for the *LAST INDEX OF* '\n'
If not found
    copy the entire data into buffer
else
    copy the data from [start to last index of '\n'] into buffer
    Then start the for loop from [last index of '\n']
endif
```

gkc (changes requested):

at_lookup/lib/src/connection/outbound_message_listener.dart
sitaram-kalluri: @gkc : The pseudo code fails in the following scenario:
For the server response: `data:@bob:phone@alice\n@alice@`, If the `data:@bob:phone@alice\n` comes in packet-1 and `@alice@` comes in packet-2. 

When the packet -1 arrives, the control goes into ELSE block and `data:@bob:phone@alice\n` goes into buffer. when packet-2 comes, the control goes into IF block and `@alice@` also goes into buffer and it keeps waiting for the terminating character resulting in time out exception.

gkc:
```
Look at the whole data and look for the *LAST INDEX OF* '\n'
If found
    Set $offset to last index of '\n'
    copy the data from [start to $offset] into buffer
Else
    Set $offset to 0
endif

Run the for loop starting from from $offset

```

gkc (approved):
I think we're good to go, here; we need a solid block of apps testing now and need to be on the lookout for situations we hadn't anticipated

PR #191, opened 2022-07-04
fix: ignore initial server response.

sitaram-kalluri:
**- What I did**
1. When new connection is established, ignore the initial server response - `@`
2. Edit the pubspec.yaml of onboarding functional test to pick the at_lookup from the path. So that the current branch at_lookup package is picked up.

**- How I did it**
1. Added an IF condition that ignores, the `@` prompt.

**- How to verify it**

### DIFF
--- a/at_lookup/lib/src/connection/outbound_message_listener.dart
+++ b/at_lookup/lib/src/connection/outbound_message_listener.dart
@@ -34,7 +34,10 @@ class OutboundMessageListener {
     var offset;
     // check buffer overflow
     _checkBufferOverFlow(data);
-
+    // When a new connection is established server returns `@` prompt.Skip it
+    if (data.length == 1 && data.first == atCharCodeUnit) {
+      return;
+    }
     // If the data contains a new line character, add until the new line char to buffer
     if (data.contains(newLineCodeUnit)) {
       offset = data.lastIndexOf(newLineCodeUnit);
--- a/at_lookup/lib/src/connection/outbound_message_listener.dart
+++ b/at_lookup/lib/src/connection/outbound_message_listener.dart
@@ -34,10 +34,6 @@ class OutboundMessageListener {
     var offset;
     // check buffer overflow
     _checkBufferOverFlow(data);
-    // When a new connection is established server returns `@` prompt.Skip it
-    if (data.length == 1 && data.first == atCharCodeUnit) {
-      return;
-    }
     // If the data contains a new line character, add until the new line char to buffer
     if (data.contains(newLineCodeUnit)) {
       offset = data.lastIndexOf(newLineCodeUnit);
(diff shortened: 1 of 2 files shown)


at_lookup/lib/src/connection/outbound_message_listener.dart
```diff
@@ -34,7 +34,10 @@ class OutboundMessageListener {
     var offset;
     // check buffer overflow
     _checkBufferOverFlow(data);
-
+    // When a new connection is established server returns `@` prompt.Skip it
+    if (data.length == 1 && data.first == atCharCodeUnit) {
+      return;
+    }
```
gkc: This isn't the right way to handle it. The problem is occurring on what is currently line 57 below, where _buffer.getData() returns an empty List, and the behaviour of `last` on an empty list is to [throw a StateError](https://api.dart.dev/stable/2.17.5/dart-core/Iterable/last.html)

So please change line 57 from
* from   `_buffer.getData().last == newLineCodeUnit`
* to        `_buffer.length() > 0 && _buffer.getData().last == newLineCodeUnit`
and remove lines 37-40 above that you added

gkc (changes requested):

at_lookup/lib/src/connection/outbound_message_listener.dart
sitaram-kalluri: Pushed the code changes.

at_onboarding_cli/functional_tests/pubspec.yaml
```diff
@@ -13,10 +13,7 @@ dependencies:
 
 dependency_overrides:
   at_lookup:
-    git:
-      url: https://github.com/atsign-foundation/at_libraries.git
-      path: at_lookup
-      ref: trunk
+    path: ../../at_lookup
```
gkc, line 16: Don't forget to remove this dependency override after this pull request has been merged to trunk

gkc (approved):

at_onboarding_cli/functional_tests/pubspec.yaml
sitaram-kalluri: @gkc: Just a thought, Can we leave dependency overrides with path so that for subsequent pushes, at_lookup will be picked from the current branch

gkc: Why?

sitaram-kalluri: Just that function tests will run the current branch instead of trunk branch. We can catch the issue when we raise PR, instead of knowing them after we merge to trunk branch

gkc: Ah sorry, yes of course that makes sense since it's all in the same repo and we're talking about the functional tests.

sitaram-kalluri: @gkc: Thank you. So leaving the dependency_overrides as is pointing to path.

PR #193, opened 2022-07-07
test: Added a test to verify core mutex behaviour

gkc:
**- What I did**
Added a test to verify core mutex behaviour

**- How I did it**
Added a test case which asserts that appropriate use of `await mutex.acquired()` and `mutex.release()` wrapping critical sections code does indeed result in expected behaviour with critical sections executing sequentially not concurrently.

**- How to verify it**
Test passes

Added a test to verify core mutex behaviour

### DIFF
--- a/at_lookup/test/mutex_test.dart
+++ b/at_lookup/test/mutex_test.dart
@@ -1,0 +1,42 @@
+import 'package:mutex/mutex.dart';
+import 'package:test/test.dart';
+
+void main() {
+  List<String> criticalSectionEvents = [];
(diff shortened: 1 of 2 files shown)


VJag (approved):
Ran it myself. LGTM.

Output:
-------
00:00 +0: Verify mutex core behaviour

criticalSection One starting

criticalSection Two starting

criticalSection One acquired mutex

criticalSection One delaying for 100 milliseconds

criticalSection One released mutex

criticalSection Two acquired mutex

criticalSection Two delaying for 10 milliseconds

criticalSection Two released mutex

00:00 +1: All tests passed!

PR #195, opened 2022-07-08
fix: make the request-response mutex apply to all AtLookupImpl instances

gkc:
**- What I did**
fix: make the mutex around request-response apply to all AtLookupImpl instances, not just a single instance 

**Note:** This is a workaround for a problem we suspect might exist in SecureSocket multiplexing

**- How I did it**
Made the `requestResponseMutex` variable static; and renamed it to `globalRequestResponseMutex`

**- How to verify it**
Should have no unexpected side-effects in application

fix: make the mutex around request-response apply to all AtLookupImpl instances, not just a single instance


gkc:
Not required

PR #197, opened 2022-07-11
feat:secure sockets with security context

### LINKED ISSUE #189: Update TLS connection to optionally output TLS keys to file
**Is your feature request related to a problem? Please describe.**
Update TLS connection to optionally output TLS keys to file. This allows you to "see" inside the TLS packets using WireShark and diagnose issues.

**Describe the solution you'd like**
Open to suggestions on implementation but it would be nice just to include a dev low level library that includes this feature. So testing can be done by including a dev library in pubspec.yaml. This would dump the TLS keys in the directory where the binary is being run.

The additional lines of code required to do this are :- 

In at_lookup_impl.dart

```
      var secConConnect = SecurityContext();
       var keyfile = File('keysfile');
       secConConnect.setTrustedCertificates('caroot/rootcacert.pem');
      var secureSocket = await SecureSocket.connect(host, int.parse(port), context: secConConnect, keyLog: (line) => keyfile.writeAsStringSync(line, mode: FileMode.append));

```

And in monitor_client.dart

```
var secConConnect = SecurityContext();
       var keyfile = File('keysfile');
       secConConnect.setTrustedCertificates('caroot/rootcacert.pem');
      var secureSocket = await SecureSocket.connect(host, int.parse(port), context: secConConnect, keyLog: (line) => keyfile.writeAsStringSync(line, mode: FileMode.append));
```

replacing the secureSocket connection with no SecurityContent()

It would be nice to abstract the SecureSocket.connect so only one change would effect both lines of code and then that abstraction could be used in the secondary server code as well perhaps.

**Describe alternatives you've considered**
I did consider pushing all the way through via command line options or by adding a method options but that I think holds the danger of leaving it in place before going to a prod build.. But open to them or other ideas.. 

**Additional context**
Screen shot of the resulting Wireshark diagnostics

image

srieteja:
closes https://github.com/atsign-foundation/at_libraries/issues/189
**- What I did**
creating secure socket with security context
provided optional feature to log TLS into a file
**- How I did it**
introduce a class SecureSocketContext which is used to create a secure socket
when "decryptPackets" flag is true, all the TLS keys are logged into a file
**- How to verify it**


at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -39,8 +39,18 @@ class AtLookupImpl implements AtLookUp {
...
+
+  String? pathToCerts;
+
+  String? tlsKeysSavePath;
+
   AtLookupImpl(String atSign, String rootDomain, int rootPort,
-      {String? privateKey, String? cramSecret}) {
+      {String? privateKey,
+      String? cramSecret,
+      this.decryptPackets,
```
murali-shris: create a class e.g 
SecurityConfig {
 bool? decryptPackets;

  String? pathToCerts;

  String? tlsKeysSavePath;
}
Pass instance of SecurityConfig to AtLookupImpl. this will reduce number of optional params passed to constructor

at_lookup/lib/src/util/secure_socket_util.dart
```diff
@@ -0,0 +1,26 @@
...
+
+  ///method that creates and returns a [SecureSocket]. If [decryptPackets] is set to true,the TLS keys are logged into a file.
+  static Future<SecureSocket> createSecureSocket(
+      String host,
+      String port,
+      bool? decryptPackets,
+      String? pathToCerts,
+      String? tlsKeysSavePath) async {
+    SecureSocketUtil.decryptPackets = decryptPackets ?? false;
+    if (SecureSocketUtil.decryptPackets) {
```
murali-shris: if(!SecureSocketUtil.decryptPackets) {
return await SecureSocket.connect(host, int.parse(port));
}
// logic for SecureSocketUtil.decryptPackets=true goes here

at_lookup/lib/src/at_lookup_impl.dart
srieteja: @murali Would introducing this in at_commons would be a good idea (or) in atLookup/at_client ?

murali-shris: at_commons

murali-shris (dismissed):

sitaram-kalluri (approved):

PR #201, opened 2022-07-13
feat: Improve timeout handling in OutboundMessageListener

murali-shris:
**- What I did**
* Replaced retries and absolute wait time with transient time out
**- How I did it**
* introduced option param transientWaitTimeMillis in read method of outbound_message_listener.dart. Added _lastReceivedTime
* when partial response is received from server and full response is not received within transientWaitTimeMillis , throw timeout exception
* when data arrives set _lastReceivedTime
* even after waiting for transientWaitTimeMillis, if full message is not received until maxWaitMillis, close connection and  throw timeoutexception
**- How to verify it**
Run the unit test group - A group of tests to verify AtTimeOutException in outbound_message_listener.dart

### DIFF
--- a/at_lookup/lib/src/connection/outbound_message_listener.dart
+++ b/at_lookup/lib/src/connection/outbound_message_listener.dart
@@ -96,7 +96,7 @@ class OutboundMessageListener {
   /// Reads the response sent by remote socket from the queue.
   /// If there is no message in queue after [maxWaitMilliSeconds], return null. Defaults to 90 seconds.
   Future<String> read(
-      {int maxWaitMillis = 90000, int transientWaitTimeMillis = 10000}) async {
+      {int maxWaitMilliSeconds = 90000, int transientWaitTimeMillis = 10000}) async {
     String result;
     _lastReceivedTime = DateTime.now();
     var startTime = DateTime.now();
--- a/at_lookup/lib/src/connection/outbound_message_listener.dart
+++ b/at_lookup/lib/src/connection/outbound_message_listener.dart
@@ -95,8 +95,11 @@ class OutboundMessageListener {
 
   /// Reads the response sent by remote socket from the queue.
   /// If there is no message in queue after [maxWaitMilliSeconds], return null. Defaults to 90 seconds.
+  /// Whenever data is received on client socket from server, [_lastReceivedTime] will be updated to current time.
+  /// [transientWaitTimeMillis] specifies the max duration to wait between current time and [_lastReceivedTime] before timing out.Defaults to 10 seconds.
   Future<String> read(
-      {int maxWaitMilliSeconds = 90000, int transientWaitTimeMillis = 10000}) async {
+      {int maxWaitMilliSeconds = 90000,
+      int transientWaitTimeMillis = 10000}) async {
     String result;
     _lastReceivedTime = DateTime.now();
     var startTime = DateTime.now();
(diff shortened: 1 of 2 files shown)


gkc (changes requested):
I’m not sure about the logic.

messageHandler is the best place to keep track of when data was last received, with read() setting an initial value of “now” for that variable. When data is received, update that variable 

Read can then do two checks 
(1) has total time exceeded max time 
(2) has difference between “now” and dataLastReceived exceeded transientWaitMillis

murali-shris:
> I’m not sure about the logic.
> 
> messageHandler is the best place to keep track of when data was last received, with read() setting an initial value of “now” for that variable. When data is received, update that variable
> 
> Read can then do two checks (1) has total time exceeded max time (2) has difference between “now” and dataLastReceived exceeded transientWaitMillis

chnages are done

murali-shris:
@gkc please review my unit tests whether they cover all the scenarios and whether the tests make sense

at_lookup/lib/src/connection/outbound_message_listener.dart
```diff
@@ -92,33 +94,46 @@ class OutboundMessageListener {
...
 
   /// Reads the response sent by remote socket from the queue.
-  /// If there is no message in queue after [maxWaitMilliSeconds], return null. Defaults to 3 seconds.
-  Future<String> read({int maxWaitMilliSeconds = 10000}) async {
-    return _read(maxWaitMillis: maxWaitMilliSeconds);
-  }
-
-  Future<String> _read({int maxWaitMillis = 10000, int retryCount = 1}) async {
+  /// If there is no message in queue after [maxWaitMilliSeconds], return null. Defaults to 90 seconds.
+  Future<String> read(
```
gkc, line 100: please add comment for transientWaitTimeMilllis parameter

at_lookup/test/outbound_message_listener_test.dart
```diff
@@ -177,4 +177,65 @@ void main() {
       expect(response, 'stream:@bob:phone@alice');
     });
   });
+
+  group('A group of tests to verify AtTimeOutException', () {
```
gkc, line 181: Good work! Main thing missing now is tests where we start a read, then we send data in chunks to the messageHandler but with delays ('await Future.delayed(Duration(xxx))`) of various lengths in between, and check both positive and negative scenarios over longer durations

gkc (changes requested):

at_lookup/lib/src/connection/outbound_message_listener.dart
murali-shris: done

at_lookup/test/outbound_message_listener_test.dart
murali-shris: added two tests
1. A test to verify full response received - delay between messages from server
2. A test to verify timeout - delay between messages from server
Please check whether this is inline with what you expected.
Will try to add more tests.

gkc: Looks good!

murali-shris:
draft PR to check at client tests with current changes in at_lookup
https://github.com/atsign-foundation/at_client_sdk/pull/605

gkc (approved):
LGTM

I see all tests passing in at_client_sdk using this branch (https://github.com/atsign-foundation/at_client_sdk/pull/605) so I think we're good to go

PR #204, opened 2022-07-19
feat: add mutex for pkam and cram authentication methods

### LINKED ISSUE #203: Race possible while authenticating
**Describe the bug**
If the `authenticate(String? privateKey)` in AtLookupImpl is called twice in quick succession, then the requests and responses for the two could become intermingled, whereas the expectation is that the entire authentication flow should complete atomically

**Expected behavior**
1. Employ a mutex, similar to how the `_process` method works
2. Within the critical section of code, add a check to begin with of the boolean which tells us whether the connection has already been authenticated should be

sitaram-kalluri:
**- What I did**
1. Add mutex around authenticate and authenticate_cram methods.

**- How I did it**

**- How to verify it**

### DIFF
--- a/at_lookup/lib/src/at_lookup_impl.dart
+++ b/at_lookup/lib/src/at_lookup_impl.dart
@@ -350,36 +350,47 @@ class AtLookupImpl implements AtLookUp {
     return await _process(atCommand, auth: auth);
   }
 
+  final Mutex _pkamAuthenticationMutex = Mutex();
+
   /// Generates digest using from verb response and [privateKey] and performs a PKAM authentication to
   /// secondary server. This method is executed for all verbs that requires authentication.
   Future<bool> authenticate(String? privateKey) async {
     if (privateKey == null) {
       throw UnAuthenticatedException('Private key not passed');
     }
-    await _sendCommand('from:$_currentAtSign\n');
-    var fromResponse = await (messageListener.read());
-    logger.finer('from result:$fromResponse');
-    if (fromResponse.isEmpty) {
-      return false;
+    try {
+      _pkamAuthenticationMutex.acquire();
+      if (!_isPkamAuthenticated) {
+        await _sendCommand('from:$_currentAtSign\n');
+        var fromResponse = await (messageListener.read());
+        logger.finer('from result:$fromResponse');
+        if (fromResponse.isEmpty) {
+          return false;
+        }
+        fromResponse = fromResponse.trim().replaceAll('data:', '');
+        logger.finer('fromResponse $fromResponse');
+        var key = RSAPrivateKey.fromString(privateKey);
+        var sha256signature =
+            key.createSHA256Signature(utf8.encode(fromResponse) as Uint8List);
+        var signature = base64Encode(sha256signature);
+        logger.finer('Sending command pkam:$signature');
+        await _sendCommand('pkam:$signature\n');
+        var pkamResponse = await messageListener.read();
+        if (pkamResponse == 'data:success') {
+          logger.info('auth success');
+          _isPkamAuthenticated = true;
+        } else {
+          throw UnAuthenticatedException(
+              'Failed connecting to $_currentAtSign. $pkamResponse');
+        }
+      }
+      return _isPkamAuthenticated;
+    } finally {
+      _pkamAuthenticationMutex.release();
     }
-    fromResponse = fromResponse.trim().replaceAll('data:', '');
-    logger.finer('fromResponse $fromResponse');
-    var key = RSAPrivateKey.fromString(privateKey);
-    var sha256signature =
-        key.createSHA256Signature(utf8.encode(fromResponse) as Uint8List);
-    var signature = base64Encode(sha256signature);
-    logger.finer('Sending command pkam:$signature');
-    await _sendCommand('pkam:$signature\n');
-    var pkamResponse = await messageListener.read();
-    if (pkamResponse == 'data:success') {
-      logger.info('auth success');
-      _isPkamAuthenticated = true;
-    } else {
-      throw UnAuthenticatedException(
-          'Failed connecting to $_currentAtSign. $pkamResponse');
-    }
-    return _isPkamAuthenticated;
   }
+
+  final Mutex _cramAuthenticationMutex = Mutex();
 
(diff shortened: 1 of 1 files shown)


at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -350,63 +350,77 @@ class AtLookupImpl implements AtLookUp {
...
-    await _sendCommand('pkam:$signature\n');
-    var pkamResponse = await messageListener.read();
-    if (pkamResponse == 'data:success') {
-      logger.info('auth success');
-      _isPkamAuthenticated = true;
-    } else {
-      throw UnAuthenticatedException(
-          'Failed connecting to $_currentAtSign. $pkamResponse');
+    try {
+      pkamAuthenticationMutex.acquire();
```
gkc: Wrap the next block of code in an if condition
```
if (!_isPkamAuthenticated) {
    await _sendCommand ..... etc
}
```

at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -350,63 +350,77 @@ class AtLookupImpl implements AtLookUp {
...
-    var digest = sha512.convert(bytes);
-    await _sendCommand('cram:$digest\n');
-    var cramResponse = await messageListener.read();
-    if (cramResponse == 'data:success') {
-      logger.info('auth success');
-      _isCramAuthenticated = true;
-    } else {
-      throw UnAuthenticatedException('Auth failed');
+    try {
+      cramAuthenticationMutex.acquire();
```
gkc: Wrap the next block of code in an if condition
```
if (!_isCramAuthenticated) {
    await _sendCommand ..... etc
}
```

gkc (changes requested):

at_lookup/lib/src/at_lookup_impl.dart
sitaram-kalluri: The next block code is moved into IF condition.

at_lookup/lib/src/at_lookup_impl.dart
sitaram-kalluri: The next block of code is moved into IF condition.

at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -350,63 +350,85 @@ class AtLookupImpl implements AtLookUp {
...
+          _isPkamAuthenticated = true;
+        } else {
+          throw UnAuthenticatedException(
+              'Failed connecting to $_currentAtSign. $pkamResponse');
+        }
+      }
+      return _isPkamAuthenticated;
+    } finally {
+      if (_pkamAuthenticationMutex.isLocked) {
+        _pkamAuthenticationMutex.release();
```
gkc: if condition not required here

at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -350,63 +350,85 @@ class AtLookupImpl implements AtLookUp {
...
+          logger.info('auth success');
+          _isCramAuthenticated = true;
+        } else {
+          throw UnAuthenticatedException('Auth failed');
+        }
+      }
+      return _isCramAuthenticated;
+    } finally {
+      if (_cramAuthenticationMutex.isLocked) {
+        _cramAuthenticationMutex.release();
```
gkc: if condition not required here

gkc (changes requested):

at_lookup/lib/src/at_lookup_impl.dart
sitaram-kalluri: The functional test in at_onboarding_cli have failed with the following error. Hence added the IF condition to check if lock is applied or not.

```
Bad state: no lock to release
  package:mutex/src/read_write_mutex.dart 164:7                         ReadWriteMutex.release
  package:mutex/src/mutex.dart 54:30                                    Mutex.release
  package:at_lookup/src/at_lookup_impl.dart 389:31                      AtLookupImpl.authenticate
  package:at_onboarding_cli/src/at_onboarding_service_impl.dart 194:28  AtOnboardingServiceImpl.authenticate
  ===== asynchronous gap ===========================
  test/at_onboarding_cli_test.dart 128:22
```

at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -350,63 +350,85 @@ class AtLookupImpl implements AtLookUp {
...
-    if (pkamResponse == 'data:success') {
-      logger.info('auth success');
-      _isPkamAuthenticated = true;
-    } else {
-      throw UnAuthenticatedException(
-          'Failed connecting to $_currentAtSign. $pkamResponse');
+    try {
+      if (!_isPkamAuthenticated) {
+        _pkamAuthenticationMutex.acquire();
+        await _sendCommand('from:$_currentAtSign\n');
```
gkc, line 364: Move `_pkamAuthenticationMutex.acquire();` to before the if statement.

at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -350,63 +350,85 @@ class AtLookupImpl implements AtLookUp {
...
-    await _sendCommand('cram:$digest\n');
-    var cramResponse = await messageListener.read();
-    if (cramResponse == 'data:success') {
-      logger.info('auth success');
-      _isCramAuthenticated = true;
-    } else {
-      throw UnAuthenticatedException('Auth failed');
+    try {
+      if (!_isCramAuthenticated) {
+        _cramAuthenticationMutex.acquire();
```
gkc: Move `_cramAuthenticationMutex.acquire();` to before the if statement.

at_lookup/lib/src/at_lookup_impl.dart
gkc: That's because I see you aren't acquiring the mutex every time. I've added comments above

gkc (changes requested):

at_lookup/lib/src/at_lookup_impl.dart
sitaram-kalluri: Got it, My bad Gary. I missed the below comments. Will make the necessary changes. Thank you.

gkc (approved):

PR #205, opened 2022-07-20
fix:modify atLookup to use Cached Secondary addresses

### LINKED ISSUE #200: at_lookup multiple calls to findSecondary
**Describe the bug**
After CacheableSecondaryAddressFinder implementation, root server should be hit only once to fetch secondary address.
But currently there are many hits to root server since findSecondary in at_lookup creates a new instance of CacheableSecondaryAddressFinder.

**To Reproduce**
1. Create a sample at_client to execute any verb
2. Enable log level to finer
3. Check for this text in the client logs
"AtLookup.findSecondary got answer:"
This will occur multiple times which implies root server is hit more than once

**Expected behavior**
Root server should be hit only once to find the secondary url. Only if the address expires from cache, subsequent connection to root server should be made

srieteja:
closes https://github.com/atsign-foundation/at_libraries/issues/200
**- What I did**
Made sure that AtLookup uses cached secondary addresses rather than calling root server each time
**- How I did it**
replaced calls to AtLookup.findSecondary to CacheableSecondaryAddressFinder.findSecondary()
**- How to verify it**

### DIFF
--- a/at_lookup/lib/src/at_lookup_impl.dart
+++ b/at_lookup/lib/src/at_lookup_impl.dart
@@ -27,6 +27,8 @@ class AtLookupImpl implements AtLookUp {
 
   OutboundConnection? get connection => _connection;
 
+  late CacheableSecondaryAddressFinder cacheableSecondaryAddressFinder;
+
   var _currentAtSign;
 
   var _rootDomain;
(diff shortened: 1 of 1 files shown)


at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -40,12 +42,16 @@ class AtLookupImpl implements AtLookUp {
   var outboundConnectionTimeout;
 
   AtLookupImpl(String atSign, String rootDomain, int rootPort,
-      {String? privateKey, String? cramSecret}) {
+      {String? privateKey,
+      String? cramSecret,
+      SecondaryAddressFinder? secondaryAddressFinder}) {
```
gkc, line 47: The findSecondary method below needs to use this secondaryAddressFinder.

gkc (changes requested):

at_lookup/lib/src/at_lookup_impl.dart
gkc: Actually, I see now that it's a static method, so ignore that. comment.

gkc (approved):

murali-shris (approved):

PR #215, opened 2022-08-02
feat: Enhance executeVerb to decode error responses in JSON.

sitaram-kalluri:
**- What I did**
- Decode the JSON responses from the server. This is in-line with the following [PR:841](https://github.com/atsign-foundation/at_server/pull/841)

**- How I did it**
- When server returns the error responses in JSON encoding, decode the response. To provide backward compatibility, on FormatException, fall back to existing implementation.

**- How to verify it**

### DIFF
--- a/at_lookup/lib/src/at_lookup_impl.dart
+++ b/at_lookup/lib/src/at_lookup_impl.dart
@@ -7,7 +7,6 @@ import 'package:at_commons/at_builders.dart';
 import 'package:at_commons/at_commons.dart';
 import 'package:at_lookup/at_lookup.dart';
 import 'package:at_lookup/src/connection/outbound_message_listener.dart';
-import 'package:at_lookup/src/util/lookup_util.dart';
 import 'package:at_utils/at_logger.dart';
 import 'package:crypto/crypto.dart';
 import 'package:crypton/crypton.dart';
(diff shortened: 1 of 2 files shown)


at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -275,12 +274,20 @@ class AtLookupImpl implements AtLookUp {
...
-        }
-        if (verbResult.split('-')[1].isNotEmpty) {
-          errorDescription = verbResult.split('-')[1];
+      try {
+        var errorMap = jsonDecode(verbResult);
+        errorCode = errorMap['errorCode'];
+        errorDescription = errorMap['errorDescription'];
+      } on FormatException {
+        // Catching the FormatException to preserve backward compatibility - responses without jsonEncoding.
+        // TODO: Can we remove the below catch block in next release once all the servers are migrated to new version.
```
gkc, line 283: We agreed that the decoding parameter should be supported in the non-JSON format as well

gkc (changes requested):

at_lookup/lib/src/at_lookup_impl.dart
sitaram-kalluri: @gkc: With reference to the following [PR-841](https://github.com/atsign-foundation/at_server/pull/841), After merging the PR-841 changes, the error messages from the server will be JSON encoded. Hence marked a TODO because server will not send the non json error responses.

gkc (approved):

PR #227, opened 2022-09-02
feat: setting socketOption tcpNoDelay -> true

srieteja:
- What I did
setting socketOption tcpNoDelay -> true while creating sockets

- How I did it
- setting the socket option in base_connection class constructor which is underlying connection for inbound and outbound connections
- setting the socket option while creating sockets from SecureSocketUtil

### DIFF
--- a/at_lookup/lib/src/connection/base_connection.dart
+++ b/at_lookup/lib/src/connection/base_connection.dart
@@ -5,12 +5,14 @@ import 'package:at_lookup/src/connection/at_connection.dart';
 
 /// Base class for common socket operations
 abstract class BaseConnection extends AtConnection {
-  final Socket _socket;
+  late final Socket _socket;
   StringBuffer? buffer;
   AtConnectionMetaData? metaData;
 
-  BaseConnection(this._socket) {
+  BaseConnection(Socket? socket) {
     buffer = StringBuffer();
+    socket?.setOption(SocketOption.tcpNoDelay, true);
+    _socket = socket!;
   }
 
   @override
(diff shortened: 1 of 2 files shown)


at_lookup/lib/src/util/secure_socket_util.dart
```diff
@@ -18,10 +21,12 @@ class SecureSocketUtil {
           throw AtException(
               'decryptPackets set to true but path to trusted certificated not provided');
         }
-        return await SecureSocket.connect(host, int.parse(port),
+        _secureSocket = await SecureSocket.connect(host, int.parse(port),
             context: securityContext,
             keyLog: (line) =>
                 keysFile.writeAsStringSync(line, mode: FileMode.append));
+        _secureSocket.setOption(SocketOption.tcpNoDelay, true);
```
VJag, line 28: Probably you have to do set option before connect

at_lookup/lib/src/connection/base_connection.dart
```diff
@@ -5,12 +5,14 @@ import 'package:at_lookup/src/connection/at_connection.dart';
...
 abstract class BaseConnection extends AtConnection {
-  final Socket _socket;
+  late final Socket _socket;
   StringBuffer? buffer;
   AtConnectionMetaData? metaData;
 
-  BaseConnection(this._socket) {
+  BaseConnection(Socket? socket) {
     buffer = StringBuffer();
+    socket?.setOption(SocketOption.tcpNoDelay, true);
```
VJag, line 14: Pls, check if we are setting this before the connect. Setting it after connect wont be effective.

at_lookup/lib/src/util/secure_socket_util.dart
srieteja: That might not be possible sir. SecureSocket class is abstract and secureSocket.connect() is a way to instantiate a secureSocket object. We can only set options after we have a secureSocket object so this is the possible way.

VJag (approved):

PR #236, opened 2022-10-09
fix: accomodate for delay in secondary creation after registering a free atsign

### LINKED ISSUE #232: Atsign auto-activation takes a while and code may progress too quickly
## Summary
Now that auto activation is a feature in the registrar, sometimes activating the atSign will take a while. Code may go straight to `onboard()` with the cram key however the secondary address is not initialized. This may become a problem if CLI code tries to onboard too quickly.

## Steps to reproduce

```
1. Activate an unactivated atSign via the web API
2. Web API will give you a cram key, but the secondary address is still null.
3. Try to onboard immediately upon receiving the cram key
4. You will receive a secondary not found because the atSign's secondary was not initialized quickly enough
```

## Hotfix

Simply pause your code until the secondary address can be found.
```dart
/// Hot fix in case activating the atSign takes too long (if the person does not press the orange "Activate" button on their atSign, then the secondary was not initialized just yet. Just run this method with async/await and it will pause your code until the secondary is successfully initialized.)
  Future<void> runUntilSecondaryExists(String rootUrl, String atSign, {timeoutIterations = 10000000}) async {
  List<String> s = rootUrl.split(':');
    String rootDomain = s[0];
    int rootPort = int.parse(s[1]);
    late SecondaryAddress sAddress;
    bool exists = false;
    int timeout = timeoutIterations;
    int count = 0;
    do {
      count++;
      try {
        sAddress = await CacheableSecondaryAddressFinder(rootDomain, rootPort).findSecondary(atSign);
        exists = true;
      } catch (e) {
        // ignore
      }

    } while(!exists && count < timeout); 
    if(sAddress.host.length > 2) {
      print('Secondary address found! ${sAddress.host}:${sAddress.port} | Iterations: $count/$timeout');
    } else {
      print('Secondary address not found after $count/$timeout iterations');
    }
    throw Exception('Secondary address not found after $count/$timeout iterations');
  }
```

### Hotfix use example:

```dart
// 1. 
// get the cram key from via the web api 
// (will actually give a non-null valid cramkey despite secondaryUrl not addressable).
String? cram = await apiUtil.authenticateAtSignValidate(atSign, otp);

// 2. 
// pause code until secondaryUrl is addressable
await apiUtil.runUntilSecondaryExists(rootUrl, atSign);

// 3. 
// onboard with the cram key (this function ultimately looks for secondaryUrl.. 
// if the pause code above were to be excluded, then it would immediately jump 
// into this code block and look for a non-existent secondaryUrl).
bool onboarded = await onboard(atSign, rootUrl, cram);
```

srieteja:
closes https://github.com/atsign-foundation/at_libraries/issues/232
**- What I did**
- introduced a method to check if secondary belonging to the atsign at hand exists or waits until the secondary is created in case it does not.

**- How I did it**
- Introduced a method called waitUntilSecondaryExists() which checks if secondary exists.
- When the secondary does not exist, try for 50 iterations to fetch the secondary address.
- Tries to connect to the secondary address fetched.
- Retries for 50 iteration if the connection to secondary was not established
- Resume normal onboarding process after abovementioned checks
**- How to verify it**

- fix: accomodate for delay in secondary creation after registering a free atsign


at_onboarding_cli/lib/src/at_onboarding_service_impl.dart
```diff
@@ -319,6 +321,36 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
     }
   }
 
+  ///Method to check if secondary belonging to [_atSign] exists
+  ///If not, wait until secondary is created
+  Future<void> _waitUntilSecondaryExists() async{
```
murali-shris: please dartfmt the code

srieteja: dart format changes pushed in latest commit

srieteja:
abandoning this PR and branch due to an issue with rebasing it with trunk code

PR #241, opened 2022-10-30
fix: add `await` when acquiring mutexes (otherwise the mutex will not work)

gkc:
**- What I did**
fix: add `await` when acquiring mutexes (otherwise the mutex will not work)

**- How to verify it**
Tests pass

fix: add `await` when acquiring mutexes (otherwise the mutex will not work)


gkc:
@kalluriramkumar added `await` when acquiring the cram and pkam mutexes

nickelskevin (approved):

PR #251, opened 2022-11-21
fix: Removed race condition related to management of outbound connection state after timeouts

### LINKED ISSUE #250: Race condition - connection may not have been closed before another connection attempt is made
### Describe the bug

When a timeout exception is thrown by the OutboundMessageListener when waiting for responses from the server, it calls async method `_closeConnection` - but without using `await`. As a result, the exception is thrown before the OutboundConnection is closed and marked as inValid. And a result of that, code which immediately tries to reissue whatever command timed out **_may_** result in AtLookupImpl's check of OutboundConnection isInvalid() returning `true`, and therefore result in AtLookupImpl's writing to a connection with a dead socket, and hanging for a very long period of time as a result.

### Steps to reproduce

Race condition, very difficult to reproduce except via unit tests, which is what #251 largely consists of, to prevent regression

### Expected behavior

After a timeout exception, the connection should be recreated before AtLookupImpl executes any other command

### Screenshots

_No response_

### Smartphones

_No response_

_No response_

### Additional context

_No response_

gkc:
fixes #250 

**- What I did**
* fix: Removed race condition related to management of OutboundConnection state after timeouts
* style: a few miscellaneous style / readability / maintainability changes
* fix: add the `unawaited_futures` lint rule to detect where we are not using `await` (usually accidentally). **Note:** This rule would have flagged the missing 'awaits' which triggered the bug 
* test: Added a new job to the github action config to run the at_lookup unit tests. Have opened a separate ticket to clean this up properly ( #252 ) - first by making the unit tests pass, then by consolidating back to single matrix job which does build and test for each package

**- How I did it**
* fix: (The fix itself) : Added missing `await`s to `OutboundMessageListener.read` which prevent problems caused (very rarely) by race condition where the OutboundConnection should have been closed but hasn't yet been closed, therefore making a rapidly-subsequent `write()` believe that the connection is still valid, and other weirdnesses to ensue
* test: Add a bunch of tests to verify
  * that isInvalid is set on the OutboundConnection after transientWaitTime timeout BEFORE the OutboundMessageListener.read() returns
  * that isInvalid is set on the OutboundConnection after maxWaitTime timeout BEFORE the OutboundMessageListener.read() returns
  * that an attempt to write to an outbound connection which has had a timeout will throw a ConnectionInvalidException
* test: Add a `@visibleForTesting Duration? delayBeforeClose;` variable to OutboundMessageListener to allow us to inject a pause before actually closing a connection. In turn makes it possible to reliably reproduce race conditions (if they exist) related to program code trying to write to a connection which should be closed but has not yet been marked as closed for whatever reason.

* fix: New lint rule flagged lots of warnings about missing awaits in test code. Added missing awaits to eliminate the warnings.
* fix: add the `unawaited_futures` lint rule to detect where we are not using `await` (usually accidentally). 
* fix: use finally block in close() to set isClosed to true on the connection metadata
* feat: log an INFO message when a new connection has been created (we already log an INFO message when we are about to try to create a new connection)
* style: remove lint warnings for spelling in test descriptions
* fix: eliminated two lint warnings by making privateKey and cramSecret initializing formal parameters in AtLookupImpl constructorImpl constructor
* fix: set transient wait time to 4 seconds and max wait time to 10 seconds for response to 'cram' verb
* fix: set transient wait time to 4 seconds and max wait time to 10 seconds for response to 'from' verb
* test: Reworded to improve clarity of mutex_test.dart output
* test: Speed up the outbound_message_listener_test unit tests by setting transientWaitTime and maxWaitTime parameters when calling `OutboundMessageListener.read()`

**- How to verify it**
* All tests pass. If you're curious and want to prove that the tests will fail without the fix, then you can remove the `await`s in outbound_message_listener.dart at lines 127 and 136 and re-run the unit tests, at which point three of them will fail.

fix: Removed race condition related to management of outbound connection state after timeouts

### DIFF
--- a/packages/at_lookup/lib/src/at_lookup_impl.dart
+++ b/packages/at_lookup/lib/src/at_lookup_impl.dart
@@ -439,7 +439,7 @@ class AtLookupImpl implements AtLookUp {
               ..atSign = _currentAtSign
               ..clientConfig = _clientConfig)
             .buildCommand());
-        var fromResponse = await messageListener.read();
+        var fromResponse = await messageListener.read(transientWaitTimeMillis: 4000, maxWaitMilliSeconds: 10000);
         logger.info('from result:$fromResponse');
         if (fromResponse.isEmpty) {
           return false;
--- a/packages/at_lookup/lib/src/at_lookup_impl.dart
+++ b/packages/at_lookup/lib/src/at_lookup_impl.dart
@@ -449,7 +449,7 @@ class AtLookupImpl implements AtLookUp {
         var bytes = utf8.encode(digestInput);
         var digest = sha512.convert(bytes);
         await _sendCommand('cram:$digest\n');
-        var cramResponse = await messageListener.read();
+        var cramResponse = await messageListener.read(transientWaitTimeMillis: 4000, maxWaitMilliSeconds: 10000);
         if (cramResponse == 'data:success') {
           logger.info('auth success');
           _connection!.getMetaData()!.isAuthenticated = true;
--- a/packages/at_lookup/lib/src/at_lookup_impl.dart
+++ b/packages/at_lookup/lib/src/at_lookup_impl.dart
@@ -229,6 +229,7 @@ class AtLookupImpl implements AtLookUp {
       //3. listen to server response
       messageListener = OutboundMessageListener(_connection!);
       messageListener.listen();
+      logger.info('New connection created OK');
     }
   }
 
--- a/packages/at_lookup/lib/src/at_lookup_impl.dart
+++ b/packages/at_lookup/lib/src/at_lookup_impl.dart
@@ -382,7 +382,7 @@ class AtLookupImpl implements AtLookUp {
 
   /// Generates digest using from verb response and [privateKey] and performs a PKAM authentication to
   /// secondary server. This method is executed for all verbs that requires authentication.
-  Future<bool> authenticatePKAM(String? privateKey) async {
+  Future<bool> authenticate(String? privateKey) async {
     if (privateKey == null) {
       throw UnAuthenticatedException('Private key not passed');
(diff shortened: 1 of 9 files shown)


cpswan (dismissed):

packages/at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -377,17 +380,15 @@ class AtLookupImpl implements AtLookUp {
     return await _process(atCommand, auth: auth);
   }
 
-  final Mutex _pkamAuthenticationMutex = Mutex();
-
   /// Generates digest using from verb response and [privateKey] and performs a PKAM authentication to
   /// secondary server. This method is executed for all verbs that requires authentication.
-  Future<bool> authenticate(String? privateKey) async {
+  Future<bool> authenticatePKAM(String? privateKey) async {
```
murali-shris: if a developer uses existing version of at_client or at_client_mobile in pub.dev and runs pub get in the app pubspec.yaml and published version of at_lookup is pulled( after this PR merge) won't the rename cause breakage?

murali-shris (changes requested):

packages/at_lookup/lib/src/at_lookup_impl.dart
gkc: Yes - I just realized that when I tested at_client_sdk against this branch of at_lookup. I have reverted the method renames

murali-shris (dismissed):

murali-shris (approved):

gkc:
[Tested at_client against this branch](https://github.com/atsign-foundation/at_client_sdk/actions/runs/3524051445) - all good. Merging to trunk.

PR #253, opened 2022-11-21
feat: At chops

murali-shris:
**- What I did**
- added a new package at_chops

**- How I did it**
- used encryption algo implementation from https://pub.dev/packages/encrypt and https://pub.dev/packages/crypton
- introduced objects for representing various keys such as AES, public private key pair,pkam key pair, cram secret in lib/src/key
- added abstract class AtEncryptionAlgorithm for encryption/decryption, AtSigningAlgorithm for data signing and verification, AtHashingAlgorithm for hashing in lib/src/algorithm/at_algorithm.dart
- RSA asymmetric encryption is implemented in lib/src/algorithm/rsa_encryption_algo.dart
- generating aes key, IV and key pairs is implemented in lib/src/util/at_chops_util.dart
- all encryption, signing and hashing methods are exposed via lib/src/at_chops_base.dart
- Pkam signing algorithm is implemented in  lib/src/algorithm/pkam_signing_algo.dart
- Signing algorithm using at_encryption key pair is implemented in lib/src/algorithm/default_signing_algo.dart
- Added AtEncryptionResult and AtEncryptionMetadata to encryption/decryption result
- Added AtSigningResult and AtSigningMetadata to signing/verification result

**- How to verify it**
* run the unit test at_chops_test


murali-shris:
@gkc @VJag 
Please review these files
packages/at_chops/lib/src/algorithm/rsa_encryption_algo.dart
packages/at_chops/lib/src/algorithm/at_algorithm.dart --> AtEncryptionAlgorithm
packages/at_chops/lib/src/key/*.dart
packages/at_chops/lib/src/at_chops_base.dart
packages/at_chops/lib/src/at_chops_impl.dart --> encryptString and decryptString

PR for at_client changes - https://github.com/atsign-foundation/at_client_sdk/pull/809/files (WIP)
I want to get an early feedback before I complete rest of the implementation in at_chops and other changes in at_client

murali-shris:
@gkc 
While instance of AtClient is created in AtClientImpl.create(...), encryption and pkam key pairs will not be available in local keystore.
Keys are persisted in at_client_service --> persistKeys which is done after at client instance creation. So I am unable to create an instance of at_chops with the keys and inject to at_client_impl
How can we proceed on this? Should I start thinking about separating persisting keys from at client instance creation?

gkc:
> @gkc While instance of AtClient is created in AtClientImpl.create(...), encryption and pkam key pairs will not be available in local keystore. Keys are persisted in at_client_service --> persistKeys which is done after at client instance creation. So I am unable to create an instance of at_chops with the keys and inject to at_client_impl How can we proceed on this? Should I start thinking about separating persisting keys from at client instance creation?

As discussed the other day - first implementation we can create it WITHIN at_client_impl if it's not been injected.

> What might a migration look like?
> Current onboarding ends up with private keys being placed into keystore
> So an initial implementation of AtChops might be done internally within AtClientImpl, as
>  .. Load private keys (and public keys) from keyStore
>  .. Create instance of AtChops with those keys
>  .. Ensure AtClient is only using AtChops for the decryption and signing operations
>  .. And ideally make sure AtClient is only using AtChops for the encryption and signature verification ops
> 
> Later ...
>   .. We need to rethink what happens with "onboarding"
>   .. once we have got keys from somewhere (atSign activation, or keychain, or a file) what do we do next?
>   .. And we need to rethink what the boostrap process is
>   .. Client program starts
>   ..   It needs to create an instance of AtChops
>   ..   Where does it get keys from?
>   ..   For a TPM / secure element, **something else** will already have put private keys in there
>   ..     And the AtChops implementation used will delegate the decryption / signing ops to the TPM
>   ..   For non-TPM (i.e. normal status quo now) where do we get the private keys from?


Once we've gotten past that first step (create instance within atclient if one's not been injected), then we can consider the following two primary scenarios:
(1) encryption and pkam key pairs have previously (e.g. in a previous run) been persisted to keystore
(2) not yet persisted because it's first time onboarding into this app

The scenario of most immediate interest is (2) - and, more specifically, (2) for non-flutter i.e. pure Dart programs. Currently we retrieve keys from a file and use at_onboarding_cli to initialize the at_client.

packages/at_chops/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+For information about how to write a good package README, see the guide for
+[writing package pages](https://dart.dev/guides/libraries/writing-package-pages). 
+
+For general information about developing packages, see the Dart guide for
+[creating packages](https://dart.dev/guides/libraries/create-library-packages)
+and the Flutter guide for
+[developing packages and plugins](https://flutter.dev/developing-packages). 
+-->
+
```
gkc, line 13: Need to update the README

gkc: Need to update the README

packages/at_chops/analysis_options.yaml
```diff
@@ -0,0 +1,30 @@
...
+# If you want a smaller set of lints you can change this to specify
+# 'package:lints/core.yaml'. These are just the most critical lints
+# (the recommended set includes the core lints).
+# The core lints are also what is used by pub.dev for scoring packages.
+
+include: package:lints/recommended.yaml
+
+# Uncomment the following section to specify additional rules.
+
+# linter:
```
gkc: Please add these lint rules
```
linter:
  rules:
    - camel_case_types
    - unnecessary_string_interpolations
    - await_only_futures
    - unawaited_futures
```

packages/at_chops/example/at_chops_example.dart
```diff
@@ -0,0 +1,4 @@
+import 'package:at_chops/at_chops.dart';
```
gkc, line 1: Are you planning to write an example of usage?

gkc (changes requested):
Really nice work Murali. Have made a few requests via comment.

Overall big picture question that occurs to me is how are we going to test forwards / backwards compatibility between clients which are or are not using AtChops. How do you plan to approach this?

packages/at_chops/example/at_chops_example.dart
murali-shris: yes Gary..in the comments I mentioned few pending items such as readme, example code

packages/at_chops/lib/src/at_chops_base.dart
```diff
@@ -0,0 +1,69 @@
...
+abstract class AtChops {
+  final AtChopsKeys _atChopsKeys;
+
+  AtChopsKeys get atChopsKeys => _atChopsKeys;
+
+  AtChops(this._atChopsKeys);
+
+  /// Encrypts the input bytes [data] using an [encryptionAlgorithm].
+  /// If [encryptionKeyType] is [EncryptionKeyType.rsa_2048] then [encryptionAlgorithm] will be set to [DefaultEncryptionAlgo]
+  Uint8List encryptBytes(Uint8List data, EncryptionKeyType encryptionKeyType,
```
gkc: As discussed previously, please can you add a 'keyName' parameter to the various asymmetric encrypt and decrypt methods which defaults to null which means legacy behaviour. 

In future we may have multiple asymmetric keypairs and we need a way to distinguish between them.

gkc (changes requested):

packages/at_chops/lib/src/at_chops_base.dart
murali-shris: multiple asymmetric keypairs - meaning different keys types e.g rsa 2048, rsa 4096 or different keys for the same key type e.g two key pairs generated for rsa 2048 ?

gkc: Both

gkc: We just need to be able to identify them. What I have in mind is that names will be derivable from the names of their public key entries in the keystore, and vice versa. For example we may store `@alice`'s public keys at
```
rsa1.__public_keys@alice
1234567.__public_keys@alice
ecc_67890_abc.__public_keys@alice
```
and then the `keyName` for each of those would be the first part i.e. 
```
rsa1
1234567
ecc_67890_abc
```

packages/at_chops/README.md
murali-shris: done

packages/at_chops/analysis_options.yaml
murali-shris: done

packages/at_chops/example/at_chops_example.dart
murali-shris: done

packages/at_chops/lib/src/at_chops_base.dart
murali-shris: done

gkc (approved):

packages/at_lookup/lib/src/at_lookup.dart
```diff
@@ -23,4 +24,7 @@ abstract class AtLookUp {
   Future<List<String>> scan({String? regex, String? sharedBy});
 
   Future<String?> executeVerb(VerbBuilder builder, {bool sync = false});
+
+  /// Set an instance of [AtChops] for signing operations during authentication
+  void setAtChops(AtChops atChops);
```
gkc: Actually, just one thing - can we make this a standard Dart setter?

gkc (changes requested):

packages/at_lookup/lib/src/at_lookup.dart
murali-shris: done

gkc (dismissed):

gkc (approved):

gkc:
@murali-shris Any reason this hasn't been merged to trunk yet?

PR #268, opened 2022-12-13
fix: issues with onboarding cli

### LINKED ISSUE #263: Register CLI - Issues
Below is a list of the issues and enhancements that have been identified while testing the most recent version of the Register CLI.

### **Issues**
**Bugs**
- [x] While running the tool from the code base, when my atSign was activating I saw multiple lines of "Connecting to secondary...." on attempt resulted in 14 lines of this and another attempt resulted in 100s of lines as this went on for a few minutes until I terminated it. 
- [x] If someone already has 10 free atSigns registered to their email and they try to get another one using the tool, it is displaying the correct message however it is then proceeding to throw an unhandled exception. Please proposed steps for handling this on page 2 of the [wording doc ](https://docs.google.com/document/d/1bM6z2Qu0XzFc-K5RYmFXqo7GE-3PYDlJ7MDIbR5hM-M/edit#) [DONE]
- [x] If someone runs the tool with an option that doesn't exist, eg -r, we are getting an unhandled exception. This should be handled in the same way as when someone runs the tool without entering any options. [DONE]
- [x] If someone runs the tool with the correct -e option but no email address, we are getting an unhandled exception. This should be handled in the same way as when someone runs the tool without enter any options. [DONE]
- [x] If someone runs the tool with the correct -e option but the email address is not in the correct format eg hhhhhhhhh, we are getting the correct message saying unable to run , you have entered an invalid email etc but then the tool proceeds to get an atSign and attempts to send the verification code to null and we get an unhandled exception. After displaying the unable to run Register Cli message, the tool should stop and prompt the person to enter their email address again [Action Required] Enter your email address: [DONE]

**Wording/Visual**
- [x] After using the tool successfully the final log displayed after letting me know where my atKeys are stored says "SEVERE|2022-12-8 10:44:12.082697|OnboardingCli}Killing current instance of at_onboarding_cli". If this is meant to signify the successful end of the process then could we replace this with something a little friendlier? SEVERE implies something has gone wrong. Maybe [Information] Your atSign Registration has been completed. 
- [x] When asked to enter my verification code, if I press enter without typing anything it says "Enter you verification code", but it should be "Enter your verification code" [DONE]
- [x] When it's activating the atSign it says "Activating you atSign: @tinashe404", but it should be "Activating your atSign: @tinashe404". This line also appears twice instead of only once.
- [x] When generating the keys it says "Generating you encryption keys and .atKeys file", but it should be "Generating your encryption keys and .atKeys file"

### **Enhancements**
- [x] Provide the Register CLI in the form of a downloadable binary. I attempted to access it using "dart pub global..." and although activation was successful, I was not able to run it. I got "Could not find an option or flag -e" after running "dart pub global at_register -e myemailaddress"
- [x] Since this tool will be used by the general public, we dont want them to have access to staging/dev and should therefore not give them the -n and --environment usage instructions [DONE]

srieteja:
closes https://github.com/atsign-foundation/at_libraries/issues/263
**- What I did**
- Fixed bug where onboarding_cli tries to connect to secondary indefinitely when the secondary is unavailable
- More argument validation in register_cli
- Minor reformatting in stdout statements register_cli and onboarding_cli
- Fixed executable usage with 'dart pub global activate'
- Upgraded dependencies

**- How I did it**
- connecting to secondary bug: missing iterator count increment, which was introduced to solve the bug
- argument validation: added exception handling in register_cli to display more understandable user messages when aruments are missing or incorrect. Replaced some exceptions with stderr and exits

- fix: minor bugs in onboarding_cli

### DIFF
--- a/packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
+++ b/packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
@@ -92,7 +92,7 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
     selfEncryptionKey = generateAESKey();
 
     stdout.writeln(
-        '[Information] Generating your encryption keys and .atKeys file\n');
+        '[Information] Generating your encryption keys and .atKeys file');
     //mapping encryption keys pairs to their names
     atKeysMap = <String, String>{
       AuthKeyType.pkamPublicKey: pkamRsaKeypair.publicKey.toString(),
@@ -193,7 +193,7 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
     logger.info(
         'atKeys file saved at ${atOnboardingPreference.downloadPath ?? atOnboardingPreference.atKeysFilePath}');
     stdout.writeln(
-        '[Success] Your .atKeys file saved at ${atOnboardingPreference.downloadPath ?? atOnboardingPreference.atKeysFilePath}\n');
+        '[Success] Your .atKeys file saved at ${atOnboardingPreference.downloadPath ?? atOnboardingPreference.atKeysFilePath}');
   }
 
   ///back-up encryption keys to local secondary
@@ -356,7 +356,7 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
     while (secureSocket == null && retryCount <= maxRetries) {
       await Future.delayed(Duration(seconds: 3));
       logger.finer('retrying connect secondary.......$retryCount/$maxRetries');
-      stdout.writeln('Connecting to secondary ...$retryCount/$maxRetries\n');
+      stdout.writeln('Connecting to secondary ...$retryCount/$maxRetries');
       try {
         secureSocket = await SecureSocket.connect(
             secondaryAddress.host, secondaryAddress.port,
--- a/packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
+++ b/packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
@@ -31,11 +31,11 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
   @override
   Future<AtClient?> getAtClient() async {
     if (_atClient == null) {
-      AtClientManager _atClientManager = AtClientManager.getInstance();
-      await _atClientManager.setCurrentAtSign(
+      AtClientManager atClientManager = AtClientManager.getInstance();
+      await atClientManager.setCurrentAtSign(
           _atSign, atOnboardingPreference.namespace, atOnboardingPreference);
-      _atLookup = _atClientManager.atClient.getRemoteSecondary()?.atLookUp;
-      return _atClientManager.atClient;
+      _atLookup = atClientManager.atClient.getRemoteSecondary()?.atLookUp;
+      return atClientManager.atClient;
     }
     return _atClient;
   }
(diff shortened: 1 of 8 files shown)


tinashe404 (dismissed):
Functionality-wise, the register_cli is doing what it needs to do. Probably need someone else to do a code review before merge.

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
```diff
@@ -23,21 +23,19 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
   AtOnboardingPreference atOnboardingPreference;
 
   AtOnboardingServiceImpl(atsign, this.atOnboardingPreference) {
-    _atSign = AtUtils.formatAtSign(atsign)!;
+    atsign = AtUtils.formatAtSign(atsign)!;
```
murali-shris, line 26: why are we changing private visibility to public here?

murali-shris: you can change this to AtUtils.fixAtSign(AtUtils.formatAtSign(atsign));
you need not change the passed parameter atsign inside the constructor

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
```diff
@@ -339,41 +329,48 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
...
+            await _atLookup?.secondaryAddressFinder.findSecondary(_atSign);
       } on Exception catch (e) {
         logger.finer(e);
       }
-      _retryCount++;
+      retryCount++;
+    }
+
+    if (secondaryAddress == null){
+      logger.severe('Could not find secondary address for $_atSign');
```
murali-shris: you can mention could not find secondary address after max retries

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
```diff
@@ -339,41 +329,48 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
...
+    retryCount = 1;
+    while (secureSocket == null && retryCount <= maxRetries) {
+      await Future.delayed(Duration(seconds: 3));
+      logger.finer('retrying connect secondary.......$retryCount/$maxRetries');
+      stdout.writeln('Connecting to secondary ...$retryCount/$maxRetries');
       try {
         secureSocket = await SecureSocket.connect(
-            _secondaryAddress!.host, _secondaryAddress.port);
+            secondaryAddress.host, secondaryAddress.port,
+            timeout: Duration(seconds: 30)); // 30-second timeout should be enough even for slow networks
```
murali-shris, line 361: check for secureSocket.remoteAddress and port !=null after SecureSocket.connect. Then set a bool flag. use the bool flag in the while loop instead of secureSocket == null

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
srieteja: addressed review comments as part of 3ecbf3191d2a90e5b21cdf9aa8b6da7421e78437

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
srieteja: addressed review comments as part of 3ecbf3191d2a90e5b21cdf9aa8b6da7421e78437

packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
srieteja: addressed review comments as part of 3ecbf3191d2a90e5b21cdf9aa8b6da7421e78437

murali-shris (approved):

PR #275, opened 2023-01-03
fix: handle cases of authentication in non-optimal network conditions

### LINKED ISSUE #271: onboardingService.authenticate(); hangs if no network connection is present
### Describe the bug

When using

    var onboard = await onboardingService.authenticate();
    
    My expected result was true if onboarded and false if the onboarding fails. But if there is for example no Internet connection then this call hangs and does not return or throw an exception. If the Internet comes back online then the onboarding seems to continue, but this await never returns true.

### Steps to reproduce

1. First I set up a VM that I can console into (so dropping the internet does not cut me off!)
2. Then switch off the connection to the Internet 
3. And then run att_talk or  similar that uses the     var onboard = await onboardingService.authenticate(); call
4. Watch the logs and see the call is hung
5. Bring back the Internet connection and see a partial recovery but the     var onboard = await onboardingService.authenticate(); never returns the future

### Expected behavior

Expected that a false would return if the onboarding fails for some reason (e.g. secondary lookup fails or network check fails). Or perhaps an exception to be thrown ? 

Either would be fine but a hang is problematic

### Screenshots

https://user-images.githubusercontent.com/6131216/208477780-db073b0f-91cf-4fb6-bce0-bb65bb7c1a73.mp4

### Smartphones

-Linux Dart
2.17.6

at_talk  - branch nosync

### Additional context

_No response_

srieteja:
closes https://github.com/atsign-foundation/at_libraries/issues/271

**- What I did**
- At the moment there is no timeout for socket.connect() calls in AtLookup. As a result of which async calls try to connect to the secondary indefinitely when the network is unavailable. By introducing a 30-sec timeout for each of these calls, when the network is unavailable the whole process times out in 1 minute and then the authenticate method returns false if the connection could not be established.

**- How I did it**
- added 30-sec timeout for socket.connect() in secondadyAddressFinder.findSecondary()
- added 30-sec timeout for socket.connect() in SecureSocketUtil.createSecureSocket()
- handle the exceptions that above-mentioned changes in case of a time-out will throw. Also, return 'false' when AtLookupImpl.authenticate() encounters an exception.
- add a network availability check in onboardingService.authenticate() that will terminate in case of network unavailability

**- How to verify it**

fix: added timeouts for socket.connect() calls

### DIFF
--- a/packages/at_lookup/lib/src/at_lookup_impl.dart
+++ b/packages/at_lookup/lib/src/at_lookup_impl.dart
@@ -386,9 +386,10 @@ class AtLookupImpl implements AtLookUp {
     if (privateKey == null) {
       throw UnAuthenticatedException('Private key not passed');
     }
-    await createConnection();
+
     try {
       await _pkamAuthenticationMutex.acquire();
+      await createConnection();
       if (!_connection!.getMetaData()!.isAuthenticated) {
         await _sendCommand((FromVerbBuilder()
               ..atSign = _currentAtSign
@@ -417,6 +418,12 @@ class AtLookupImpl implements AtLookUp {
         }
       }
       return _connection!.getMetaData()!.isAuthenticated;
+    } on Exception catch (e) {
+      logger.severe(e.toString());
+      return false;
+    } catch (e) {
+      logger.severe(e.toString());
+      return false;
     } finally {
       _pkamAuthenticationMutex.release();
     }
@@ -586,7 +593,7 @@ class AtLookupImpl implements AtLookUp {
   }
 
   Future<void> close() async {
-    await _connection!.close();
+    await _connection?.close();
   }
 
   Future<void> _sendCommand(String command) async {
--- a/packages/at_lookup/lib/src/at_lookup_impl.dart
+++ b/packages/at_lookup/lib/src/at_lookup_impl.dart
@@ -421,9 +421,6 @@ class AtLookupImpl implements AtLookUp {
     } on Exception catch (e) {
       logger.severe(e.toString());
       return false;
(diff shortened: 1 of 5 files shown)


packages/at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -386,9 +386,10 @@ class AtLookupImpl implements AtLookUp {
     if (privateKey == null) {
       throw UnAuthenticatedException('Private key not passed');
     }
-    await createConnection();
+
     try {
       await _pkamAuthenticationMutex.acquire();
+      await createConnection();
```
murali-shris, line 392: should the createConnection be before mutex acquire?

srieteja: If the create connection is before the mutex acquire and the createConnection() encounters an exception; it will execute the finally block where we are trying to release a mutex that hasn't been acquired yet. Alternatively, we can put createConnection() in a separate try-catch block where it can return false upon encountering an exception. This way we can have the mutex acquire after the createConnection().

packages/at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -417,6 +418,9 @@ class AtLookupImpl implements AtLookUp {
         }
       }
       return _connection!.getMetaData()!.isAuthenticated;
+    } on Exception catch (e) {
```
gkc, line 421: This is a potentially breaking change - when UnAuthenticatedException is thrown, it will now be immediately caught by this catch block and not thrown back to the caller as it would have previously.

PR #306, opened 2023-03-17
feat: at_chops changes for pkam using private key from secure element

murali-shris:
**- What I did**
 - changes to support pkam using private key from secure element
 - analyser changes
 - bug fix in encoding/decoding of signature
**- How I did it**
- introduced a public method readPublicKey(String publicKeyId) in at_chops_base. Clients using secure element will be extending AtChopsImpl and override readPublicKey(...) method
- provided setters for encryption key pair and symmetric key. If these keys are not available during at_chops creation, they can be set later on.
- added base64Encoding to the recently added method sign(..). Modified documentation for base64encoding/base64decoding
**- How to verify it**
n/a since no changes in implementation in at_chops

### DIFF
--- a/packages/at_chops/lib/src/algorithm/ecc_signing_algo.dart
+++ b/packages/at_chops/lib/src/algorithm/ecc_signing_algo.dart
@@ -30,7 +30,7 @@ class EccSigningAlgo implements AtSigningAlgorithm {
   bool verify(Uint8List signedData, Uint8List signature, {String? publicKey}) {
     if (publicKey == null && _privateKey == null) {
       throw AtSigningVerificationException(
-          'public key not set not ecc verification');
+          'Cannot verify signature - publicKey not provided or privateKey not available');
     }
     var ec = elliptic.getSecp256r1();
     var hashHex = sha256.convert(signedData).toString();
--- a/packages/at_chops/lib/src/algorithm/ecc_signing_algo.dart
+++ b/packages/at_chops/lib/src/algorithm/ecc_signing_algo.dart
@@ -29,7 +29,8 @@ class EccSigningAlgo implements AtSigningAlgorithm {
   @override
   bool verify(Uint8List signedData, Uint8List signature, {String? publicKey}) {
     if (publicKey == null && _privateKey == null) {
-      throw AtSigningVerificationException('public key not set not ecc verification');
+      throw AtSigningVerificationException(
+          'public key not set not ecc verification');
     }
     var ec = elliptic.getSecp256r1();
     var hashHex = sha256.convert(signedData).toString();
--- a/packages/at_chops/lib/src/at_chops_base.dart
+++ b/packages/at_chops/lib/src/at_chops_base.dart
@@ -111,7 +111,8 @@ abstract class AtChops {
   /// Refer to [DefaultHash] for default implementation of hashing.
   String hash(Uint8List signedData, AtHashingAlgorithm hashingAlgorithm);
 
-  /// Reads a public key from SIM card based on ID. SIM card should have an IOTSafe compliant app
-  /// with a key pair preinstalled.
+  /// Reads a public key from a secure element or any other source
+  /// Clients can implement this method if want to use a different keypair for pkam authentication.
+  /// e.g. Secure element can be a SIM card with an IOTSafe compliant app with a key pair preinstalled.
   String readPublicKey(String publicKeyId);
 }
--- a/packages/at_chops/lib/src/at_chops_base.dart
+++ b/packages/at_chops/lib/src/at_chops_base.dart
@@ -76,9 +76,9 @@ abstract class AtChops {
   // ignore: deprecated_member_use_from_same_package
   /// If [signingKeyType] is [SigningKeyType.signingSha256] then [signingAlgorithm] will be set to [DefaultSigningAlgo]
   AtSigningResult verifySignatureBytes(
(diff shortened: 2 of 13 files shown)


packages/at_chops/lib/src/at_chops_base.dart
```diff
@@ -108,4 +110,8 @@ abstract class AtChops {
   /// Create a hash of input [signedData] using a [hashingAlgorithm].
   /// Refer to [DefaultHash] for default implementation of hashing.
   String hash(Uint8List signedData, AtHashingAlgorithm hashingAlgorithm);
+
+  /// Reads a public key from SIM card based on ID. SIM card should have an IOTSafe compliant app
```
gkc: From SIM is only one example of what an implementation could do.

packages/at_chops/lib/src/at_chops_impl.dart
```diff
@@ -361,4 +361,10 @@ class AtChopsImpl extends AtChops {
       throw InvalidDataException('Unrecognized type of data: $data');
     }
   }
+
+  @override
+  String readPublicKey(String publicKeyId) {
+    // This method is implemented only for extensions of AtChops that use secure element for private keys
```
gkc: Again this is just one example use (the main example for sure) but I could choose to implement this in some other way also

gkc (changes requested):
Docs could be made more precise

packages/at_chops/lib/src/at_chops_base.dart
murali-shris: please check whether my latest docs are fine.

packages/at_chops/lib/src/at_chops_impl.dart
murali-shris: please check whether my latest docs are fine.

packages/at_chops/lib/src/algorithm/ecc_signing_algo.dart
```diff
@@ -29,7 +29,8 @@ class EccSigningAlgo implements AtSigningAlgorithm {
   @override
   bool verify(Uint8List signedData, Uint8List signature, {String? publicKey}) {
     if (publicKey == null && _privateKey == null) {
-      throw AtSigningVerificationException('public key not set not ecc verification');
+      throw AtSigningVerificationException(
+          'public key not set not ecc verification');
```
gkc: 1. I don't really understand the exception message here. Should it be, "Cannot verify signature - publicKey not provided or privateKey not available"
2. Should it be an `||` condition rather than an `&&`?

packages/at_chops/lib/src/at_chops_impl.dart
```diff
@@ -364,7 +364,7 @@ class AtChopsImpl extends AtChops {
 
   @override
   String readPublicKey(String publicKeyId) {
-    // This method is implemented only for extensions of AtChops that use secure element for private keys
+    // This method is implemented only for extensions of AtChops that use secure element or any other source for private keys apart from the default source(.atKeys file)
```
gkc: `other than` rather than `apart from`

gkc (changes requested):

murali-shris:
Found a bug while testing. We replaced at_chops.signString(..) method with at_chops.sign(..) in at_lookup(yet to be merged). signString was base64 encoding the pkam signature..That is missing in sign method. Will fix this bug in this PR.

packages/at_chops/lib/src/algorithm/ecc_signing_algo.dart
murali-shris: In ECDSA, public key can be derived from the private key.
E.g pubKey = _privateKey!.publicKey;
There are two ways verify methods can be called.
1) final eccAlgo = EccSigningAlgo();
eccAlgo.privateKey = //some ecc private key;
privateKey will be used for signing.While verifying signatue, publickey will be derived from the private key and used for verification
2) Zariot use case.Private is not available to us. 
So we will invoke EccSigningAlgo().verify(data, signature, //pass the public key)
I am throwing an exception if both are not available to verify method. if it is || then it will always be true. Hope my explanation is correct?

packages/at_chops/lib/src/at_chops_impl.dart
murali-shris: changed.

murali-shris:
> Found a bug while testing. We replaced at_chops.signString(..) method with at_chops.sign(..) in at_lookup(yet to be merged). signString was base64 encoding the pkam signature..That is missing in sign method. Will fix this bug in this PR.

Fixed the issue. Ran the changes with at_client and at server
https://github.com/atsign-foundation/at_client_sdk/pull/965
https://github.com/atsign-foundation/at_server/pull/1267 (end2test failure due to test flakiness - not related to at_chops changes)

packages/at_chops/lib/src/algorithm/ecc_signing_algo.dart
gkc: Got it - makes sense. Can you update the exception message as I suggested please?

gkc (changes requested):
LGTM except for an exception message that needs to be made clearer

packages/at_chops/lib/src/algorithm/ecc_signing_algo.dart
murali-shris: done

gkc (approved):
LGTM

PR #310, opened 2023-03-18
feat: updating home_directory to the correct location

nickelskevin:
**- What I did**
updated atctivate_cli.dart to use  "$homeDirectory/.atsign/keys";
**- How I did it**

**- How to verify it**

feat: updating home_direcrtory to  "$homeDirectory/.atsign/keys";

### DIFF
--- a/packages/at_onboarding_cli/lib/src/activate_cli/activate_cli.dart
+++ b/packages/at_onboarding_cli/lib/src/activate_cli/activate_cli.dart
@@ -4,6 +4,8 @@ import 'dart:io';
 
 import 'package:at_utils/at_logger.dart';
 
+import '../home_directory/home_directory.dart';
+
 Future<void> main(List<String> arguments) async {
   //defaults
   String rootServer = 'root.atsign.org';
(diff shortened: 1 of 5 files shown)


packages/at_onboarding_cli/lib/src/activate_cli/activate_cli.dart
```diff
@@ -41,7 +43,10 @@ Future<void> main(List<String> arguments) async {
 
   stdout.writeln('[Information] Root server is ${argResults['rootServer']}');
 
-  var downloadPath = '${Directory.current.path}/keys';
+  // var downloadPath = '${Directory.current.path}/keys';
+  var homeDirectory = getHomeDirectory();
+  var downloadPath = "$homeDirectory/.atsign/keys";
```
gkc, line 48: Do you need to use platform-specific path separators? e.g. \ for Windows etc

gkc (commented):
Are you intending a release? If so, please bump version in pubspec, and update CHANGELOG

packages/at_onboarding_cli/lib/src/activate_cli/activate_cli.dart
cconstab: We use the same code in attalk on windows. I really need to put those into a package..

cconstab (dismissed):
Thank you

packages/at_onboarding_cli/lib/src/activate_cli/activate_cli.dart
cconstab: Updated ready for v1.2.5 release to pub.dev

cconstab (approved):
checked and tested manually

PR #313, opened 2023-03-20
feat:created dockerfile to create at_activate

cconstab:
Created Dockerfile to be able to produce at_activate cli using docker and buildx

### DIFF
--- a/packages/at_onboarding_cli/Dockerfile
+++ b/packages/at_onboarding_cli/Dockerfile
@@ -14,7 +14,7 @@ RUN \
   dart pub get ; \
   dart pub upgrade ; \
   dart compile exe bin/activate_cli.dart -o $BINARYDIR/at_activate ; \
-  tar cvzf /atsign/at_activate.`uname -m`.tar.gz at_activate/at_activate
+  tar cvzf /atsign/at_activate."$(uname -m)".tgz "$BINARYDIR"/at_activate
 
   
 # Second stage of build FROM scratch
--- a/packages/at_onboarding_cli/Dockerfile
+++ b/packages/at_onboarding_cli/Dockerfile
@@ -1,0 +1,22 @@
+FROM dart:2.19.4@sha256:5d2f42239d719bef7fe60abb37273aea16d8526980662b696c218663ac005d96 AS buildimage
+ENV HOMEDIR=/atsign
+ENV BINARYDIR=/atsign/at_activate
+ENV USER_ID=1024
+ENV GROUP_ID=1024
+WORKDIR /atsign
+# Context for this Dockerfile needs to be at_server repo packages/at_root_server
+# If building manually then (from packages/at_root_server):
+## sudo docker build -t atsigncompany/root .
+COPY . .
+RUN \
+  mkdir -p $HOMEDIR ; \
+  mkdir -p $BINARYDIR ; \
+  dart pub get ; \
+  dart pub upgrade ; \
+  dart compile exe bin/activate_cli.dart -o $BINARYDIR/at_activate ; \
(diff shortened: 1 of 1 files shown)


gkc:
Works for me. Made the comments accurate for this repo.

Would the next step be to create github workflows (similar to those in at_server's at_server.yaml) which automatically build and push images to dockerhub?
- dev images, upon merge to trunk
- prod images, upon creation of a release

packages/at_onboarding_cli/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# If building manually then (from packages/at_onboarding_cli):
+## docker build -t atsigncompany/at_activate .
+COPY . .
+RUN \
+  mkdir -p $HOMEDIR ; \
+  mkdir -p $BINARYDIR ; \
+  dart pub get ; \
+  dart pub upgrade ; \
+  dart compile exe bin/activate_cli.dart -o $BINARYDIR/at_activate ; \
+  tar cvzf /atsign/at_activate.`uname -m`.tar.gz at_activate/at_activate
```
cpswan: A few nits here:

* Backticks for shell commands are discouraged in style guides, so maybe use `"$(uname -m)"`
* We could simplify what's going to be a complex filename by having a `.tgz` suffix (as these days people aren't going to separately gunzip then untar).
* `at_activate/at_activate` assumes our working directory is /atsign (which it is) but we've got something of a random mix of `$DIR` and `/dir` and `dir` and it would be nice to have a more consistent style.

So altogether this line could (should?) be:
`tar cvzf /atsign/at_activate."$(uname -m)".tgz "$BINARYDIR"/at_activate`

cpswan (commented):
I'm not clear why we're creating a Docker image with a tarfile inside it? Why not just the AOT binary (and runtime underneath it) so that somebody could then `docker run atsigncompany/at_activate <params>`?

We could then have a multi platform build, though that's only going to work for things that the official Dart Docker image supports (Linux flavours, so no Windows or Mac).

packages/at_onboarding_cli/Dockerfile
cconstab: This is primarily for downloads and releases to go alongside sshnp which follows a similar pattern of tar.gz. with no support for windows.
Using docker buildx this docker file can create some , and in the future maybe all of them with darwin/ .
Doing this by hand is tedious so this helps a little.

cconstab: @cpswan I tried using the style guide "$uname -m)" and it does not work with or without the " so have gone back to backticks..

thanks for the other edits

cpswan: Was it failing because of the missing opening bracket, or is that just a typo?

cpswan:
If we need to automate creation of the tar files then that can just be a GitHub action workflow rather than using Docker. I'll take a pass at doing one in the UK morning.

packages/at_onboarding_cli/Dockerfile
cconstab: just a typo I tested with your edit ...

cconstab:
Thanks Chris.. If you do get it working then we can use the same pattern to create sshnp which would be very cool

cpswan:
@cconstab I think we're going to need this (in addition to anything we might do with GitHub Actions matrix builds).

I've made some changes to the Dockerfile to align with my proposal at https://github.com/atsign-foundation/sshnoports/issues/106

LMK if you're OK with it, and I'll merge.

cpswan (approved):

cconstab:
Thanks Chris this looks great.. We should do the same thing for sshnoports repoas  it saves yak shaving

PR #344, opened 2023-05-11
build: Changes for Dart 3

gkc:
**- What I did**
**Note:** No errors flagged by Dart 3 in code; none of the packages here _have_ to be republished; however at_contacts probably ought to be
build: Modify max dart version in pubspecs to <4.0.0
build: update minimum dart SDK version to 2.15.0 in order to pick up the `unawaited` function
feat: removed usages of deprecated `formatAtSign` method from at_contacts_impl.dart
style: changed several for loops to change `for (var foo in <collection>)` to `for (final foo in <collection>)`
build: Updated package version and CHANGELOG for at_contact in prep for publishing a new version

### DIFF
--- a/packages/at_contact/lib/src/at_contacts_impl.dart
+++ b/packages/at_contact/lib/src/at_contacts_impl.dart
@@ -150,7 +150,7 @@ class AtContactsImpl implements AtContactsLibrary {
     if (scanList.isEmpty) {
       return contactList;
     }
-    for (var key in scanList) {
+    for (final key in scanList) {
       var atsign = reduceKey(key.key!);
       var atKey = _formAtKeyFromScanKeys(key);
       AtContact? contact;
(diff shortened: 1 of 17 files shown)


packages/at_contact/lib/src/at_contacts_impl.dart
```diff
@@ -418,11 +416,11 @@ class AtContactsImpl implements AtContactsLibrary {
     }
     var atKey = _formKey(KeyType.group, key: atGroup.groupId!);
     // Add all contacts in atContacts from atGroup
-    atContacts.forEach((contact) {
+    for (var contact in atContacts) {
```
cpswan: could be `final`

cpswan (dismissed):

packages/at_contact/lib/src/at_contacts_impl.dart
gkc: Changed. Changed a few other for loops also.

cpswan (approved):

PR #358, opened 2023-06-09
feat: initial commit of at_cli_commons package

gkc:
**- What I did**
feat: initial commit of at_cli_commons package

**- How I did it**
- New package with basic README
- A subsequent commit will add a usage example of CLIBase to the README before we publish to pub.dev

### DIFF
--- a/packages/at_cli_commons/lib/src/home_directory.dart
+++ b/packages/at_cli_commons/lib/src/home_directory.dart
@@ -1,0 +1,23 @@
+import 'dart:io';
+
+// Get the home directory or null if unknown.
+String? getHomeDirectory() {
+  switch (Platform.operatingSystem) {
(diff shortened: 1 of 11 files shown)


packages/at_cli_commons/lib/src/home_directory.dart
```diff
@@ -0,0 +1,23 @@
+import 'dart:io';
+
+// Get the home directory or null if unknown.
+String? getHomeDirectory() {
+  switch (Platform.operatingSystem) {
+    case 'linux':
```
cpswan, line 6: missing return?

cconstab: No need for one

cconstab (approved):

PR #374, opened 2023-07-05
fix: Remove LoggingType and add LoggingHandler as optional argument t…

### LINKED ISSUE #343: Ability to log to stderr **by default**
### Is your feature request related to a problem? Please describe.

Recently we added the ability to specific a different output handler for AtSignLogger. However, this must be set for each AtSignLogger constructed. This means it is not possible to make all AtSignLoggers used by your program send their log to stderr - since most of the loggers are created in dependent libraries

### Describe the solution you'd like

- Make a static method on AtSignLogger to allow setting of a default log output handler
- modify AtSignLogger constructor to allow nulls for the output handler, and if null then use the current static default

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

sitaram-kalluri:
**- What I did**
* Replace the `LoggingType` enum in `AtSignLogger` constructor with `LoggingHandler` to enable custom logging.
* In AtSignLogger, add static variable `defaultLoggingHandler` and default it to instance of `ConsoleLoggingHandler`
* Export `handlers.dart` file in at_loggers.dart for custom logging classes to extend `LoggingHandler` class
* Resolve dart analyze issues in `atsign_util.dart` and `fix_at_sign_test.dart`
* Remove the code related to `LoggingType` enum and commented code.
* Update the version of lints and test packages

### DIFF
--- a/packages/at_utils/lib/src/logging/atsignlogger.dart
+++ b/packages/at_utils/lib/src/logging/atsignlogger.dart
@@ -11,13 +11,15 @@ class AtSignLogger {
   static String _root_level = 'info';
   bool _hierarchicalLoggingEnabled = false;
 
-  static final ConsoleLoggingHandler _consoleLoggingHandler =
+  static final ConsoleLoggingHandler consoleLoggingHandler =
       ConsoleLoggingHandler();
+  static final StdErrLoggingHandler stdErrLoggingHandler =
+      StdErrLoggingHandler();
 
   /// The default logging handler to log events.
   ///
   /// Defaults to [ConsoleLoggingHandler] which writes log events to console.
-  static LoggingHandler defaultLoggingHandler = _consoleLoggingHandler;
+  static LoggingHandler defaultLoggingHandler = consoleLoggingHandler;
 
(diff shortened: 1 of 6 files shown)


packages/at_utils/lib/src/logging/atsignlogger.dart
```diff
@@ -13,19 +13,38 @@ class AtSignLogger {
 
   static final ConsoleLoggingHandler _consoleLoggingHandler =
       ConsoleLoggingHandler();
-  static final StdErrLoggingHandler _stdErrLoggingHandler =
-      StdErrLoggingHandler();
+
```
gkc, line 18: Make _consoleLoggingHandler public, as it is final anyway
Undelete _stdErrLoggingHandler (but make it public also)

Then, while it's starting up, application code can do `AtSignLogger.defaultLoggingHandler = AtSignLogger.stdErrLoggingHandler;`

gkc (changes requested):
LGTM, couple of minor changes requested

gkc (approved):
LGTM! Thanks @sitaram-kalluri

PR #389, opened 2023-07-24
fix: do not store pkamPublicKey, encryptionPublicKey on LocalSecondary

### LINKED ISSUE #1075: at_client expects internal keys to have namespace (when namespace is not provided through preferences)
### Describe the bug

some atsigns after being activated with activate_cli/onboarding_cli end up in a situation where sync keeps failing due to an exception 
````
SEVERE|2023-06-26 13:34:51.888723|SyncService (@34thundering)|Exception in sync cb049548-7780-4cf0-95c1-7bde8faa1d93. Reason: Failed to syncData caused by namespace is mandatory 
````

When a namespace is not set through at_client preferences, at_client while creating internal keys (e.g. lastReceivedCommitId, lastReceivedNotification) expects a namespace explicitly defined for these keys. And since these keys do not have a specific namespace, the client throws an exception which terminates the client.

### Steps to reproduce

1. First I 'activate a new atsign with activate_cli'
2. Then I 'use that atsign with onboaring_cli'
3. And then 'insert a key into the local secondary'
4. This results in a Sync Failure and the syncs keep failing forever

### Expected behavior

Sync should succeed and not throw any exceptions

### Possible solution

1. exclude internal keys from the namespace scrutiny (possibly in the validatePutRequest() in at_client_impl)
2. use a constant/predefined namespace for all internal keys

AtOnbaordingCli

srieteja:
partially resolves https://github.com/atsign-foundation/at_client_sdk/issues/1075

**- What I did**
- Resolve sync conflict caused by encryptionPublicKey
- Make namespace mandatory for authentication

**- How I did it**
- do not insert pkamPublicKey and encryptionPublicKey into LocalSecondary
- throw an exception when namespace is not provided during authentication

**- How to verify it**
- test yet to come

- Make namespace mandatory for authentication

### DIFF
--- a/packages/at_onboarding_cli/CHANGELOG.md
+++ b/packages/at_onboarding_cli/CHANGELOG.md
@@ -1,5 +1,7 @@
+## 1.3.1
+- feat: make namespace mandatory for authentication
 ## 1.3.0
-- feat: Introduced verification-code based activation of atsigns
+- feat: Introduced verification-code based activation of atsign's
 - fix: deprecate qr_code based activation
 - feat: introduced new exceptions
 - fix: improve existing logger messages and added some
--- a/packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
+++ b/packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
@@ -61,7 +61,7 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
       logger.severe(
           'Please set namespace in AtOnboardingPreferences.'
               ' Cannot create at_client instance without namespace.');
-      throw AtOnboardingException('namespace cannot be null');
+      throw AtOnboardingException('namespace cannot be empty');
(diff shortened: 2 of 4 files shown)


murali-shris:
@srieteja Please revisit these changes after we complete client side changes for APKAM

srieteja:
> @srieteja Please revisit these changes after we complete client side changes for APKAM

Will do @murali-shris

gkc:
> > @srieteja Please revisit these changes after we complete client side changes for APKAM
> 
> Will do @murali-shris

@srieteja @murali-shris  Is it time to revisit this PR?

srieteja:
This was supposed to be a solution for a bug we had in the past, but we found a better solution. This is not relevant anymore.

gkc:
> This was supposed to be a solution for a bug we had in the past, but we found a better solution. This is not relevant anymore.

Excellent, thanks @srieteja

PR #391, opened 2023-08-03
fix: changes to enroll syntax

sitaram-kalluri:
**- What I did**
- Modify enroll verb to accept a json for params. Added new class in at_commons EnrollParams.dart
- Rename the enrollApprovalId to enrollmentId for naming consistency
- Add "list" to "EnrollOperationEnum"
- Update the unit tests to for the new enroll verb syntax
- Export "enroll_params.dart" file


packages/at_commons/lib/src/verb/syntax.dart
```diff
@@ -128,14 +128,10 @@ class VerbSyntax {
...
-      // Add colon only when operation is request, approve, deny or revoke. Do not add for list
-      r'(?:(?<=request|approve|deny|revoke):)?'
-      r'(?:(?:enrollmentId:(?<enrollmentId>[a-zA-Z0-9-]+))|'
-      r'(?:appName:(?<appName>[a-zA-Z0-9]+):)?'
-      r'(?:deviceName:(?<deviceName>[a-zA-Z0-9_]+):)?'
-      r'(?:namespaces:\[(?<namespaces>[a-zA-Z0-9;_,]+)\]:)?'
-      r'(?:totp:(?<totp>[0-9]+):)?'
-      r'(?:apkamPublicKey:(?<apkamPublicKey>.*)))?$';
+      // Positive lookbehind - Matches the value of operation named group
+      //  To ignore ":" if operation is list
```
gkc: To be precise, this is actually to require ":" if operation is request|approve|deny|revoke

packages/at_commons/lib/src/verb/syntax.dart
```diff
@@ -130,6 +130,8 @@ class VerbSyntax {
       r'^enroll:(?<operation>request|approve|deny|revoke|list)'
       // Positive lookbehind - Matches the value of operation named group
       //  To ignore ":" if operation is list
+      //  To be precise, this is actually to require ":"
+      //  if operation is request|approve|deny|revoke
```
gkc: I didn't mean for you to literally transcribe what I wrote :-)

How about, `Look-behind to require ':' if operations is request|approve|deny|revoke`

On looking at this again, shouldn't we also require params when operation is request|approve|deny|revoke ?

An alternative regex would be to require `enroll:` then require either `list$` or `(request|approve|deny|revoke):(params fragment)$` which might be simpler to maintain

gkc (changes requested):

packages/at_commons/lib/src/verb/syntax.dart
sitaram-kalluri: Is the following regex looks good?
```^enroll:(?<operation>(list$)|(request|approve|deny|revoke|):)(?<enrollParams>.+)?$```

gkc: I think you need to write some unit tests

sitaram-kalluri: We have unit tests in "packages/at_commons/test/syntax_test.dart" to verify the enroll verb syntax. Will test the changes and add more tests to it.

gkc (approved):

PR #393, opened 2023-08-17
feat: at_chops apkam changes

murali-shris:
**- What I did**
- apkam changes in at_chops package
**- How I did it**
- Deprecated symmetricKey in AtChopsKeys and introduced selfEncryptionKey and apkamSymmetricKey to clearly differentiate between these two symmetric keys
- Upgraded at_commons and at_utils in pubspec. Removed at_onboarding_cli dependency since it creates circular dependency between at_chops and at_onboarding_cli
- Removed onboard_secure_element.dart from at_chop example. This file will be moved to onboarding_cli example
- Updated pubspec version and modified changelog for publishing at_chops
**- How to verify it**
- n/a


packages/at_chops/lib/src/key/impl/at_chops_keys.dart
```diff
@@ -4,14 +4,29 @@ import 'package:at_chops/src/key/impl/at_pkam_key_pair.dart';
...
   AtPkamKeyPair? _atPkamKeyPair;
+
+  /// Key pair for data signing and verification
   AtSigningKeyPair? atSigningKeyPair;
+
+  @Deprecated('Use selfEncryptionKey')
   SymmetricKey? symmetricKey;
 
+  /// Default self encryption key
+  SymmetricKey? selfEncryptionKey;
```
gkc, line 23: If the intent is that this replaces `symmetricKey` then shouldn't `symmetricKey` instance variable be replaced by a setter and getter for `symmetricKey` which use the `selfEncryptionKey` instance variable instead?

gkc (changes requested):

packages/at_chops/lib/src/key/impl/at_chops_keys.dart
murali-shris: my intent is to differentiate instance variable between self encryption symmetric key and apkam symmetric key.. If I keep a setter/getter symmetricKey, then the caller will be confused whether this is self encryption symmetric key or apkam symmetric key.

gkc: Sorry my note wasn't clear.

You are adding `selfEncryptionKey`, which is "the same as" the current `symmetricKey`. You are marking `symmetricKey` as deprecated, but you are not removing it. As a result you have two instance variables which mean the same thing. But instead of keeping `symmetricKey` as is you could do this:
```
@Deprecated('Use selfEncryptionKey')
SymmetricKey? get symmetricKey => selfEncryptionKey;
@Deprecated('Use selfEncryptionKey')
void set symmetricKey(SymmetricKey? sk) => selfEncryptionKey = sk;
```

gkc (approved):

PR #401, opened 2023-09-03
feat: make atServer address lookup resilient to transient failures to reach atDirectory

### LINKED ISSUE #368: address lookup requests to atDirectory (root) service should retry at least once upon failure
### Describe the bug

We should retry at least once if root server is unavailable at lookup time, so that the secondary address lookup operation is resilient to any transient issues / partial failures of root service.

**Note:** While this ticket is in at_libraries because that's where the at_lookup package is, this resilience to degraded root service availability needs to be added for both clients and atServers. `at_secondary_server` I believe has its own implementation of secondary address finding. The problem needs to be resolved there also. Ideally we could take this opportunity to change at_server so that, instead of having its own impl, it also uses the at_lookup package to do lookups on root

### Additional context

See [this post-mortem](https://docs.google.com/document/d/1mX3XtSajT3L7I9ou67SOMOg1GV17dsWF1gm55tbRoao) on a recent incident.

gkc:
closes #368 

**- What I did**
- feat: Added retry behaviour to `SecondaryUrlFinder`. If root lookup fails with an exception, the lookup will be retried up to four times, with default delays of 50, 100, 150 and 200 after each failure.
- feat: made `retryDelaysMillis` a public static variable in `SecondaryUrlFinder`; this allows clients to control (1) how many retries are done and (2) the delay after each subsequent retry

**- How I did it**
Note to reviewers: the 'real' changes are in `cacheable_secondary_address_finder.dart` and `secondary_address_cache_test.dart`. Changes to other test files are just the results of refactoring of some shared test code.

- modified `SecondaryUrlFinder` so that it will retry up to 4 (number of retries is configurable) times with delays of [50, 100, 150, 200] milliseconds respectively (these delays are configurable) in the event of some exception while creating a connection to the atDirectory
- modified `SecondaryUrlFinder` to allow injection of a socket factory, so we can test in unit tests
- add unit tests for atDirectory retry behaviour
  - full lookup on a non-existent atDirectory
  - lookup on a mocked atDirectory with 0, 1, 2, 3, 4 or 5 simulated failures
- docs: added some docs for SecondaryUrlFinder
- feat: made `retryDelaysMillis` a public static varaiable in SecondaryUrlFinder; this allows clients to control (1) how many retries are done and (2) the delay after each subsequent retry
- refactor unit tests
  - extract shared mock definitions into new file at_lookup_test_utils.dart
  - remove duplicates of the `createMockSecureSocket` function and moved into at_lookup_test_utils.dart
- refactor: more refactoring and tidying of shared test code

**- How to verify it**
- [x] Existing and new at_lookup tests pass
- at_client_sdk and at_server tests pass when using this at_lookup branch
  - [x] [at_client_sdk](https://github.com/atsign-foundation/at_client_sdk/pull/1115)
  - [x] [at_server](https://github.com/atsign-foundation/at_server/pull/1528)

### DIFF
--- a/packages/at_lookup/CHANGELOG.md
+++ b/packages/at_lookup/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 3.0.40
+- feat: make atServer address lookup resilient to transient failures to reach
+  atDirectory
 ## 3.0.39
 - feat: Changes for apkam
 - chore: Upgraded at_commons to 3.0.53 and at_utils to 3.0.15
--- a/packages/at_lookup/CHANGELOG.md
+++ b/packages/at_lookup/CHANGELOG.md
@@ -1,6 +1,10 @@
 ## 3.0.40
-- feat: make atServer address lookup resilient to transient failures to reach
-  atDirectory
+- feat: make `SecondaryUrlFinder` (atServer address lookup) resilient to 
+  transient failures to reach an atDirectory
+- feat: made `retryDelaysMillis` a public static variable
+  in `SecondaryUrlFinder`; this allows clients to control
+  - (1) how many retries are done and
+  - (2) the delay after each subsequent retry
 ## 3.0.39
 - feat: Changes for apkam
 - chore: Upgraded at_commons to 3.0.53 and at_utils to 3.0.15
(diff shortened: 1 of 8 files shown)


cconstab (approved):
Looks great...TY

PR #431, opened 2023-10-17
fix: Replace use of deprecated variables

sitaram-kalluri:
**- What I did**
- In at_commons package, replaced the usage of deprecated variables with new variables in "AtConstants"


gkc (approved):
Thank you very much @sitaram-kalluri

PR #441, opened 2023-10-26
feat: APKAM onboarding changes

murali-shris:
**- What I did**
- apkam changes for at_onboarding_cli
**- How I did it**
- Introduced enroll method in AtOnboardingService. Clients/apps can used this method to place an enrollment request
- Added enrollmentId as an optional param to authenticate method. After enrollment, clients must pass enrollmentId for apkam auth
- onboarding changes. 
    - added a flag enableEnrollmentDuringOnboard=false in AtOnboardingPreference. Clients/apps needing privileges to approve enrollments have to set this flag to true 
    - onboard method will thrown an exception if enableEnrollmentDuringOnboard=true and appName or deviceName is not set in OnboardingPreference. Both the params are required for enrolling the app on server
    - onboarding logic is delegated to AtAuth --> onboard method
- enroll method impl
    -  generate new apkam key pair and apkam symmetric key
    - retrieve encryption private key from server and use it to encrypt  apkam symmetric key
    - send enroll request to server with appName, deviceName,otp, encrypted apkam symmetric key and apkam public key from apkam key pair
    - asynchronously attempt apkam to server until enrollment is approved or denied
    - once apkam auth is success then add enrollmentId to _pkamSuccessController. This will trigger the method _listenToPkamSuccessStream which will save the security keys to .atKeys file

- _activateAtsign changes
    - encrypt default encryption key and self encryption key with first client's apkam symmetric key.
    - send enroll request to server which gets auto approved
    - if enroll request is approved, attempt pkam with enrollmentId. if pkam is success then update encryption public key to server, delete cram secret, generate keys file, persist keys to local secondary 
- changes to local hive storage path
    - if hive storage path is not in set in preferences, default path will include enrollmentId (if available). This will ensure two clients enrolled from same machine do not have the same hive storage path.

**- How to verify it**
- run the functional tests in at_onboarding_cli
- run the examples for APKAM in example/. Check README.md in example/

### DIFF
--- a/packages/at_onboarding_cli/example/README.md
+++ b/packages/at_onboarding_cli/example/README.md
@@ -1,0 +1,29 @@
+List of steps to run the examples for checking apkam enrollment
+
+1. Onboard an atsign which has privilege to approve/deny enrollments
+   dart example/onboard.dart <atsign> <path_store_keys_file>
+   e.g. dart example/onboard.dart @alice🛠 /home/alice/.atsign/@alice🛠_wavikey.atKeys
+2. Authenticate using the onboarded atsign
+   dart example/apkam_authenticate.dart <atsign> <path_of_keys_file_from_#1>
+   e.g. dart example/apkam_authenticate.dart @alice🛠 /home/alice/.atsign/@alice🛠_wavikey.atKeys
+3. Run client to approve enrollments
+   dart example/enroll_app_listen.dart <atsign> <path_of_keys_file_from_#1>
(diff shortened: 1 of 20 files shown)


packages/at_onboarding_cli/example/README.md
```diff
@@ -0,0 +1,29 @@
+List of steps to run the examples for checking apkam enrollment
```
murali-shris, line 1: I have requested @srieteja to go over this readme and test the flow. Replace normal args with ArgParser and usage

gkc (approved):
LGTM thank you @murali-shris

PR #451, opened 2023-11-15
fix: removed private key reference and from at_lookup_impl

### LINKED ISSUE #434: remove privateKey references from at_lookup_impl and enforce at_chops
### Describe the bug

After at_chops and at_auth package implementation, private key should no longer be read from preferences and used in code.
Some code was left behind in atLookUp as a fallback mechanism if atChops is not set.
Changes to enforce atChops and remove privateKey from test files has been done in 
https://github.com/atsign-foundation/at_client_sdk/pull/1142

### Steps to reproduce

n/a

### Expected behavior

* Remove call to authenticate method in AtLookupImpl --> _process
else if (privateKey != null) {
          logger.finer('calling pkam without atchops');
          await authenticate(privateKey);
        } 
* Test the changes with at_client/at_client_mobile/at_onboarding_cli branches.
* Do a sanity with mobile/CLI apps

### Screenshots

_No response_

### Smartphones

_No response_

_No response_

### Additional context

_No response_

purnimavenkatasubbu:
**- What I did**
- Removed privateKey reference in the at_lookup_impl
- Remove piece of code that used a private key class variable to perform auth

**- How to verify it**
- Manual testing has been done with at_client/at_client_mobile and all the apps 
- existing unit/functional/e2e tests passing should count as verification

fix: removed private key reference and from at_lookup_impl


packages/at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -34,7 +34,6 @@ class AtLookupImpl implements AtLookUp {
 
   late int _rootPort;
 
-  String? privateKey;
```
murali-shris, line 37: mark this as deprecated instead of removing.if any old clients set private key,  they won't break.

murali-shris (changes requested):

packages/at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -55,7 +54,7 @@ class AtLookupImpl implements AtLookUp {
   AtChops? _atChops;
 
   AtLookupImpl(String atSign, String rootDomain, int rootPort,
-      {this.privateKey,
```
murali-shris: keep privateKey in constructor for backward compatibility

packages/at_lookup/lib/src/at_lookup_impl.dart
purnimavenkatasubbu: updated

packages/at_lookup/lib/src/at_lookup_impl.dart
purnimavenkatasubbu: updated

murali-shris (approved):

packages/at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -573,14 +574,11 @@ class AtLookupImpl implements AtLookUp {
         if (_atChops != null) {
           logger.finer('calling pkam using atchops');
           await pkamAuthenticate(enrollmentId: enrollmentId);
-        } else if (privateKey != null) {
-          logger.finer('calling pkam without atchops');
-          await authenticate(privateKey);
```
gkc, line 578: The above lines of code should not have been removed @sitaram-kalluri

PR #455, opened 2023-11-23
feat: Add put operation to OTP syntax for SPP

sitaram-kalluri:
**- What I did**
- Enhance OTP verb syntax to accept a pass code and store in the secondary server.

**- How I did it**
- To the existing OTP verb regex, add 'put' to the operation named group, which then accepts 6 digit alpha-numeric characters.

**- How to verify it**
- Added unit tests to verify the OTP verb syntax.

- Add "put" operation to OTP verb to store semi-permanent pass codes

NOTE: Following is commit for the corresponding changes in the secondary server. Will raise PR on publishing at_commons package: https://github.com/atsign-foundation/at_server/compare/trunk...semi-permanent-pass-codes

### DIFF
--- a/packages/at_commons/CHANGELOG.md
+++ b/packages/at_commons/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 3.0.58
+- fix: Add "put" operation to OTP verb to store semi-permanent pass codes
 ## 3.0.57
 - feat: Introduced TTL(Time to Live) for OTP verb to configure OTP expiry
 ## 3.0.56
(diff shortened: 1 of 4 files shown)


gkc (dismissed):

gkc (approved):

gkc:
@sitaram-kalluri In the atServer, please implement so that (1) there is only one semi-permanent PIN - i.e. if you `put` again it overwrites the previous one (2) check the privileges of the client connection's enrollment - only connections with privileged access (access to the _manage namespace) should be able to do an `otp:put`

sitaram-kalluri:
> @sitaram-kalluri In the atServer, please implement so that (1) there is only one semi-permanent PIN - i.e. if you `put` again it overwrites the previous one (2) check the privileges of the client connection's enrollment - only connections with privileged access (access to the _manage namespace) should be able to do an `otp:put`

@gkc: Sure.

PR #463, opened 2023-11-27
fix: Remove deprecated methods and fields and rename enrollStatus field.

sitaram-kalluri:
**- What I did**
- Since the at_commons version is updated to v4.0.0., removed the deprecated methods
- Renamed 'EnrollStatus" as "EnrollmentStatus".
- Removed "EnrollResponse" which currently does not have usage. Instead "AtEnrollmentResponse" is used which is defined in at_auth package.
- Run code formatter


gkc:
@sitaram-kalluri Checks are failing in at_commons

gkc (changes requested):
Checks are failing in at_commons

gkc (approved):
LGTM

PR #466, opened 2023-12-01
fix: at_cli_commons: handle malformed atsigns (no leading `@`) in CLIBase constructor

gkc:
**- What I did**
fix: handle malformed atsigns (no leading `@`) in CLIBase constructor

**- How I did it**
See commits


cconstab (dismissed):
Is atChops defaulted now hence the removal?

gkc:
> Is atChops defaulted now hence the removal?

@cconstab  Yes deprecated in at_client 3.0.68 - but thanks for spotting that, I've now updated the pubspec also to require at_client ^3.0.68 ... can you re-review please?

cconstab (approved):

PR #471, opened 2023-12-07
fix: Remove metadata fields in verb builders

sitaram-kalluri:
**- What I did**
- Remove AtKey fields and metadata fields from the verb builders. Use AtKey instance in "AbstractVerbBuilder" to set the key.
- Since "key" in AtKey cannot be null, removed "?" which indicates that key can be null, and marked it as late.
- Removed `static const String atOperation = 'operation` from "at_constants.dart" file because it is redundant.
- Remove "?" for "isBinary" and "isEncrypted" fields default them to false.
- In AtClient pacakge,  AtKey.Metadata is verififed for null check and if null assigned to `Metadata()` . Initialized Metadata field to remove the null check in at_client methods.
- Since metadata is removed from verb builders, removed `packages/at_commons/lib/src/verb/metadata_using_verb_builder.dart` file which is not needed. 
- Uptake the at_commons changes into at_lookup.
- Modify the tests in at_commons.

Raised draft PR's in at_client_sdk and at_server to validate the changes:

at_server: https://github.com/atsign-foundation/at_server/pull/1688

at_client: https://github.com/atsign-foundation/at_client_sdk/pull/1179


packages/at_lookup/lib/src/at_lookup_impl.dart
```diff
@@ -94,10 +94,11 @@ class AtLookupImpl implements AtLookUp {
   Future<bool> delete(String key,
       {String? sharedWith, bool isPublic = false}) async {
     var builder = DeleteVerbBuilder()
```
murali-shris: at_lookup changes you can move to separate PR.
at_lookup dependency on at_commons 4.x is a minor version release for at_lookup?

murali-shris:
in verb builders replace string concatenation with string builder

murali-shris (changes requested):

sitaram-kalluri:
> in verb builders replace string concatenation with string builder

This change is completed and pushed the changes.

packages/at_lookup/lib/src/at_lookup_impl.dart
sitaram-kalluri: - Sure Murali, WIll move the at_lookup changes to a new PR.
- At_lookup can pusblished as a minor version because we are not introducing any breaking changes.

murali-shris (dismissed):

murali-shris (approved):

PR #507, opened 2024-01-29
feat: Modify "executeCommand" to throw exception on error response from atServer (as "executeVerb" does)

sitaram-kalluri:
**- What I did**
- Currently, the server response are only parsed in the "executeVerb" method. But the requests that sent via the "executeCommand" are not parsed. The responses are returned as is to the client.
- Modify the "executeCommand" method to parse the server response and then return the response or the exception.

**- How I did**
- Move the logic that parses the server response in the "executeVerb" method to a private method "_verbResponseHandler". 
- Move the logic that parses the error response to a private method "_errorResponseHandler". The "_verbResponseHandler" invokes the "_errorResponseHandler" if the server response starts with "error:".
- Pass the response received via the execute command to _verbResponseHandler to parse the server response.

### DIFF
--- a/packages/at_lookup/CHANGELOG.md
+++ b/packages/at_lookup/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 3.0.46
+- fix: Modify "executeCommand" to parse the error response from server and return appropriate exception
 ## 3.0.45
 - build[deps]: Upgraded at_chops to v2.0.0
 ## 3.0.44
--- a/packages/at_lookup/lib/src/at_lookup_impl.dart
+++ b/packages/at_lookup/lib/src/at_lookup_impl.dart
@@ -326,10 +326,22 @@ class AtLookupImpl implements AtLookUp {
     // Setting the errorCode and errorDescription to default values.
     var errorCode = 'AT0014';
     var errorDescription = 'Unknown server error';
-
-    var errorMap = jsonDecode(verbResult);
-    errorCode = errorMap['errorCode'];
-    errorDescription = errorMap['errorDescription'];
+    try {
+      var errorMap = jsonDecode(verbResult);
+      errorCode = errorMap['errorCode'];
+      errorDescription = errorMap['errorDescription'];
+    } on FormatException {
+      // Catching the FormatException to preserve backward compatibility - responses without jsonEncoding.
+      // TODO: Can we remove the below catch block in next release once all the servers are migrated to new version.
+      if (verbResult.contains('-')) {
+        if (verbResult.split('-')[0].isNotEmpty) {
+          errorCode = verbResult.split('-')[0];
+        }
+        if (verbResult.split('-')[1].isNotEmpty) {
+          errorDescription = verbResult.split('-')[1];
+        }
+      }
+    }
 
     throw AtLookUpException(errorCode, errorDescription);
   }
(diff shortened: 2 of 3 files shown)


gkc (changes requested):
1. Why have you removed the backwards compatibility?
2. Is this just intended as a refactoring? Because it seems like the existing behaviour is exactly what you are describing (parse the error response and throw an exception)

sitaram-kalluri:
> 1. Why have you removed the backwards compatibility?
> 2. Is this just intended as a refactoring? Because it seems like the existing behaviour is exactly what you are describing (parse the error response and throw an exception)

@gkc : Since there is a TODO to remove the code in next release and the client versions above v3.0.37 (currently 3.0.75) are capable of handling the JSON encoding the server error responses, it is mostly unlikely to encounter the format exception. Hence removed the code. However, since this is not a major release, please let me know if wanted me to put back the code related to backward compatibility.

gkc:
> > 1. Why have you removed the backwards compatibility?
> 
> > 2. Is this just intended as a refactoring? Because it seems like the existing behaviour is exactly what you are describing (parse the error response and throw an exception)
> 
> 
> 
> @gkc : Since there is a TODO to remove the code in next release and the client versions above v3.0.37 (currently 3.0.75) are capable of handling the JSON encoding the server error responses, it is mostly unlikely to encounter the format exception. Hence removed the code. However, since this is not a major release, please let me know if wanted me to put back the code related to backward compatibility.

Ok - so the PR is not a fix, it's removing unnecessary code which is there for compatibility with old atServers?

sitaram-kalluri:
> > > 1. Why have you removed the backwards compatibility?
> > 
> > 
> > > 2. Is this just intended as a refactoring? Because it seems like the existing behaviour is exactly what you are describing (parse the error response and throw an exception)
> > 
> > 
> > @gkc : Since there is a TODO to remove the code in next release and the client versions above v3.0.37 (currently 3.0.75) are capable of handling the JSON encoding the server error responses, it is mostly unlikely to encounter the format exception. Hence removed the code. However, since this is not a major release, please let me know if wanted me to put back the code related to backward compatibility.
> 
> Ok - so the PR is not a fix, it's removing unnecessary code which is there for compatibility with old atServers?

@gkc : Yes, But, just not introduce any breaking change, I am adding the changes related to backward compatibility , will defer this to later.

gkc:
please do the backwards compatibility change also. What confused me a lot here is that the PR title says it's a fix, when it is not. Please update the PR's title and description to correctly reflect its purpose

sitaram-kalluri:
> please do the backwards compatibility change also. What confused me a lot here is that the PR title says it's a fix, when it is not. Please update the PR's title and description to correctly reflect its purpose

Sure, Gary. Added back the backward compatibility change. Now the PR contains only the fix that handles the error responses from server in the executeCommand command to return the exception.

gkc:
Please update the PR's title and description to correctly reflect its purpose

gkc:
The PR description says:

> - Add verb response handler to handle the server responses.
>   - Add "_verbResponseHandler" to handle the server responses
>   - If response starts with "data:" then return the response.
>   - If the response starts with "error:" parse the exception and return the appropriate exception.

That's the existing functionality. But the PR is not adding anything new; it is removing some code which has become unnecessary over time.

gkc (commented):
LGTM. Please confirm that at_client tests are successful against this branch

sitaram-kalluri:
> LGTM. Please confirm that at_client tests are successful against this branch

@gkc : Ran tests in at_client against this branch and the runs are successful.

gkc (approved):

PR #511, opened 2024-02-03
feat:replace md5 checksum

murali-shris:
**- What I did**
- at_commons changes to replace md5 with sha256/512
**- How I did it**
- introduced new object PublicKeyHash which has the hash of public key and hashing algo used.
- added custom object PublicKeyHash to at_key.dart
- changes to toJson and toProtocolFragment methods in metadata
- deprecated pubKeyCS in at_key.dart
- added unit tests

**- How to verify it**
- unit tests should pass

### DIFF
--- a/packages/at_commons/lib/src/keystore/public_key_hash.dart
+++ b/packages/at_commons/lib/src/keystore/public_key_hash.dart
@@ -17,7 +17,10 @@ class PublicKeyHash {
     return map;
   }
 
-  static PublicKeyHash fromJson(Map json) {
+  static PublicKeyHash? fromJson(Map? json) {
+    if (json == null) {
+      return null;
+    }
     return PublicKeyHash(
         json['hash'], PublicKeyHashingAlgo.values.byName(json['algo']));
   }
--- a/packages/at_commons/lib/src/keystore/public_key_hash.dart
+++ b/packages/at_commons/lib/src/keystore/public_key_hash.dart
@@ -1,0 +1,19 @@
+/// Represents hash of an atsign's public encryption key and the hashing algorithm used
+class PublicKeyHash {
+  String? hash;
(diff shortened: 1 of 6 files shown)


packages/at_commons/lib/src/keystore/public_key_hash.dart
```diff
@@ -0,0 +1,30 @@
+/// Represents hash of an atsign's public encryption key and the hashing algorithm used
+class PublicKeyHash {
+  String? hash;
+  PublicKeyHashingAlgo? publicKeyHashingAlgo;
```
gkc: Why are these nullable?

murali-shris: field should not be nullable. I have removed nullable fields

gkc (approved):

gkc:
@murali-shris pubKeyCS is marked as deprecated but pubKeyHash is not being used in any other package that I can find. What's the plan here? I think to prevent confusion we need to remove the deprecation annotation until at_server and at_client_sdk fully support pubKeyHash. I'll make a PR to that effect. (PR [here](https://github.com/atsign-foundation/at_libraries/pull/615) was merged a while back)

gkc:
@murali-shris Please see above comment; can you create a ticket to uptake pubKeyHash in (a) atServer (b) at_client_sdk (c) at_c_sdk

murali-shris:
> @murali-shris Please see above comment; can you create a ticket to uptake pubKeyHash in (a) atServer (b) at_client_sdk (c) at_c_sdk
https://github.com/atsign-foundation/at_server/issues/2121
https://github.com/atsign-foundation/at_client_sdk/issues/1417
https://github.com/atsign-foundation/at_c/issues/424

PR #522, opened 2024-02-22
fix: Update the examples as per the changes in at_commons 4.0.0

sitaram-kalluri:
**- What I did**
- With upgrading at_commons to 4.x.x, the examples in at_lookup are obsolete and throws compile errors .
- Update the examples in at_lookup as per the changes in at_commons 4.x.x


gkc (approved):
LGTM

PR #525, opened 2024-02-26
fix: Add ignore to fix dart analyze issue

sitaram-kalluri:
**- What I did**
- Fix dart analyze issues in at_commons
- When running `dart pub publish --dry-run` observed the below message, which says to use at_commons-4.0.2 (which is retracted version).  Hence setting back the at_commons version to 4.0.2

```
Total compressed archive size: 59 KB.
Validating package... (2.3s)
Package validation found the following hint:
* The previous version is 4.0.1.

  It seems you are not publishing an incremental update.

  Consider one of:
  * 5.0.0 for a breaking release.
  * 4.1.0 for a minor release.
  * 4.0.2 for a patch release.

Package has 0 warnings and 1 hint.
The server may enforce additional checks.
```

### DIFF
--- a/packages/at_commons/CHANGELOG.md
+++ b/packages/at_commons/CHANGELOG.md
@@ -1,6 +1,7 @@
+## 4.0.3
+- fix: "toJson()" invoked on "pubKeyHash" leads to NullPointerException.
 ## 4.0.2
 - feat: changes to replace md5 checksum - deprecated pubKeyCS in AtKey and introduced new class PublicKeyHash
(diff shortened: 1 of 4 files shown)


gkc (changes requested):
@sitaram-kalluri I don't think you can re-publish 4.0.2

sitaram-kalluri:
> @sitaram-kalluri I don't think you can re-publish 4.0.2

@gkc: Alright, I will update the version to 4.0.3 and push the changes.

gkc (approved):

PR #535, opened 2024-03-21
fix: at_cli_commons: CLIBase: Write to stderr, not stdout

gkc:
**- What I did**
fix: CLIBase: Write to stderr, not stdout

**- How to verify it**
tests pass

### DIFF
--- a/packages/at_cli_commons/lib/src/cli_base.dart
+++ b/packages/at_cli_commons/lib/src/cli_base.dart
@@ -142,7 +142,8 @@ class CLIBase {
         atKeysFilePath ?? '$homeDir/.atsign/keys/${this.atSign}_key.atKeys';
     localStoragePathToUse =
         storageDir ?? '$homeDir/.$nameSpace/${this.atSign}/storage';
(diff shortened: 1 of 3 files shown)


packages/at_cli_commons/lib/src/cli_base.dart
```diff
@@ -142,7 +142,8 @@ class CLIBase {
         atKeysFilePath ?? '$homeDir/.atsign/keys/${this.atSign}_key.atKeys';
     localStoragePathToUse =
         storageDir ?? '$homeDir/.$nameSpace/${this.atSign}/storage';
-    downloadPathToUse = downloadDir ?? '$homeDir/.$nameSpace/${this.atSign}/files';
+    downloadPathToUse =
+        downloadDir ?? '$homeDir/.$nameSpace/${this.atSign}/files';
```
gkc, line 146: this was dart format

xavierchanth (approved):
Ideally (later on) we make it so you can pass in an io sink to push progress messages to.

PR #530, opened 2024-03-13
feat: update enroll regex to accept params and new EnrollParamsField introduction

srieteja:
**- What I did**
- [feat: update enroll regex -> enroll:list accepts optional params](https://github.com/atsign-foundation/at_libraries/commit/a2176021ea3c53f87ed2cfc9ad485cd4480d28ee)
- [feat: add new field in EnrollParams named 'approvalStatusFilter'](https://github.com/atsign-foundation/at_libraries/commit/512e1de63fae7254f61747fab4a1ea4315bda1ad)

**- Description for the changelog**

feat: update enroll regex to accept params and new EnrollParamsField introduction

PR #536, opened 2024-03-22
fix: backwards compatability issue with new enroll verb syntax

srieteja:
**- What I did**
- Remove default values for enrollmentStatusFilter  from EnrollParams and EnrollVerbBuilder

**- More context**
- Following changes were made in 'https://github.com/atsign-foundation/at_libraries/pull/530':
> 1. Enroll verb syntax was updated to accept verbParams for enroll list (which previously was not the case)
> 2. Another change made in EnrollParams and EnrollVerbBuilder defaulted 'enrollmentStatusFilter' to EnrollmentStatus.values (all values) which when used by the EnrollVerbBuilder.buildCommand() would change the default enroll list command to '_enroll:list:{"enrollmentStatusFilter":["pending","approved","denied","revoked","expired"]}_'.

- Now this change in syntax and verb builder was being consumed by the client immediately, but not the server. This mismatch caused the server to throw and InvalidSyntax exception.

**- How this PR fixes things**
- Changes in this PR will change the default enroll list command to match the syntax that the AtServer accepts, hence the client although working with a NEW EnrollVerbSyntax will send out a command that matches the OLD EnrollVerbSyntax

**- How to verify it**
- The following is a draft PR in at_client_sdk that consumes the changes in this PR and runs the github test suite against it: https://github.com/atsign-foundation/at_client_sdk/pull/1271

fix: Remove default values for enrollmentStatusFilter from EnrollParams and EnrollVerbBuilder

### DIFF
--- a/packages/at_commons/CHANGELOG.md
+++ b/packages/at_commons/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 4.0.4
+## 4.0.5
 - feat: Enhance enroll:list to enable filtering based on enrollment status
 ## 4.0.3
 - fix: "toJson()" invoked on "pubKeyHash" leads to NullPointerException.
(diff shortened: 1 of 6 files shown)


sitaram-kalluri:
We can have the default values. I believe merging the changes in at_secondary_server to trunk will resolve the issue related to at_client PR, than doing another release of at_commons

murali-shris (approved):

PR #590, opened 2024-05-29
fix: Add revoke to at_enrollment_base.dart

sitaram-kalluri:
**- What I did**
- Add "revoke" method to "AtEnrollmentBase" to enable clients to revoke an approved enrollment. The revoke method accepts an instance of "EnrollmentRequestDescision" and instance of "AtLookup" as input parameters to perform revoke operation.
- Added a static method "revoked" to "EnrollmentRequestDecision" to get an instance of "EnrollmentRequestDecision".
- Update the version in pubspec.yaml and CHANGELOG.md

### DIFF
--- a/packages/at_auth/lib/src/enroll/enrollment_request_decision.dart
+++ b/packages/at_auth/lib/src/enroll/enrollment_request_decision.dart
@@ -84,7 +84,7 @@ class EnrollmentRequestDecision {
     return enrollmentRequestDecision;
   }
 
-  /// If the request is denied, the requester application is prevented from logging into the application.
+  /// If the request is denied, the requester application is prevented from authenticating to the atServer.
   ///
   /// ```dart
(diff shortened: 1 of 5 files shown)


packages/at_auth/lib/src/enroll/enrollment_request_decision.dart
```diff
@@ -75,7 +84,7 @@ class EnrollmentRequestDecision {
     return enrollmentRequestDecision;
   }
 
-  /// If the request is denied, the requester is prevented from logging into the application.
+  /// If the request is denied, the requester application is prevented from logging into the application.
```
gkc: "is prevented from authenticating to the atServer"

gkc (commented):
One minor doc change please otherwise LGTM

packages/at_auth/lib/src/enroll/enrollment_request_decision.dart
sitaram-kalluri: Updated the dart docs

gkc (approved):

PR #654, opened 2024-09-12
feat: add delete option to enroll verb

srieteja:
**- What I did**
- introduce option 'delete' in enrollVerbBuilder and enroll verb regex
- remove update operation from EnrollVerb regex and EnrollOperationEnum 

**- How to verify it**
- unit tests present to validate the changes

feat: add delete option to enroll verb

### DIFF
--- a/packages/at_commons/lib/src/verb/syntax.dart
+++ b/packages/at_commons/lib/src/verb/syntax.dart
@@ -128,7 +128,7 @@ class VerbSyntax {
   static const notifyRemove = r'notify:remove:(?<id>[\w\d\-\_]+)';
   static const enroll =
       // The non-capturing group (?::)? matches ":" if the operation is request|approve|deny|revoke
-      r'^enroll:(?<operation>(?:(request|approve|deny|revoke|update|list|fetch|unrevoke)))(:(?<force>force))?(?::)?((?<enrollParams>.+)|(<=list:)<enrollParams>.?)?$';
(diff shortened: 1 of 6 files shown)


packages/at_commons/lib/src/verb/syntax.dart
```diff
@@ -128,7 +128,7 @@ class VerbSyntax {
   static const notifyRemove = r'notify:remove:(?<id>[\w\d\-\_]+)';
   static const enroll =
       // The non-capturing group (?::)? matches ":" if the operation is request|approve|deny|revoke
-      r'^enroll:(?<operation>(?:(request|approve|deny|revoke|update|list|fetch|unrevoke)))(:(?<force>force))?(?::)?((?<enrollParams>.+)|(<=list:)<enrollParams>.?)?$';
+      r'^enroll:(?<operation>(?:(request|approve|deny|revoke|list|fetch|unrevoke|delete)))(:(?<force>force))?(?::)?((?<enrollParams>.+)|(<=list:)<enrollParams>.?)?$';
```
murali-shris, line 131: why is enroll:update removed?

murali-shris (changes requested):

packages/at_commons/lib/src/verb/syntax.dart
srieteja: Update was previously added by me when working on enroll:update. But we decided not to implement enroll:update. Looks like it was left behind from back then, it has no functionality.

srieteja: It was also left behind in EnrollOperationEnum, that has also been removed in [52d34e4](https://github.com/atsign-foundation/at_libraries/pull/654/commits/52d34e41a639bea8b5f8147428d101ba1912388f)

murali-shris (approved):

PR #669, opened 2024-09-24
feat: at_commons changes for setting isEncrypted for both true and false

murali-shris:
**- What I did**
- changes for setting isEncrypted in metadata for update verb for both true and false. Previously isEncrypted is set only when true
**- How I did it**
- modified atKey.metadata.toProtocolFragment
- removed adhoc code for notify
- modified test assertions
**- How to verify it**
- tests should pass

### DIFF
--- a/packages/at_commons/CHANGELOG.md
+++ b/packages/at_commons/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## 5.0.0
-- [Breaking Change]feat: Add isEncrypted flag to update verb metadata for both true and false
+- [Breaking Change]feat: Emit the isEncrypted value in the metadata if it is false
 - fix: update pkam regex to accept sha512 as hashing algo
 ## 4.1.2
 - feat: Add "expiry" enroll params to support apkam keys to auto expiry after specified time duration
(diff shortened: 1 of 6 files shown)


gkc (changes requested):
- Given that `isEncrypted` defaults to false today
- Then an app built with an older version of at_client but this version of at_commons will start emitting 'isEncrypted:false' in the metadata even though the data will actually be encrypted
- Can you think through the ramifications of this please, @murali-shris - do we need an at_commons major version change?

murali-shris:
> * Given that `isEncrypted` defaults to false today
> * Then an app built with an older version of at_client but this version of at_commons will start emitting 'isEncrypted:false' in the metadata even though the data will actually be encrypted
> * Can you think through the ramifications of this please, @murali-shris - do we need an at_commons major version change?

Right Gary. Older version of atClient with latest at_commons version will have encrypted data with isEncrypted false.
Impact of this will be similar to existing data for shared/self keys. There is no functional breakage but will end up with more incorrect metadata(data is encrypted but isEncrypted =false)
I guess it is better to have a major release for this change.

packages/at_commons/CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
-## 4.1.3
+## 5.0.0
+- [Breaking Change]feat: Add isEncrypted flag to update verb metadata for both true and false
```
gkc: Change to "Emit the `isEncrypted` value in the metadata if it is false"

murali-shris: Done

gkc (approved):

PR #690, opened 2024-10-14
fix: Check of write permission of atKeys file path when submitting enroll request

### LINKED ISSUE #668: ensure atKeys file is writeable before finalizing pkam / apkam
### Describe the bug

- set permissions such that atKeys file isn't writeable
- submit enrollment
- apkam file creation fails
- no atKeys file! What now?

### Steps to reproduce

.

### Expected behavior

pkam / apkam flow should complete only if atKeys file has been written successfully

### Screenshots

_No response_

### Smartphones

_No response_

_No response_

### Additional context

_No response_

sitaram-kalluri:
**- What I did**
- When submitting the enrollment request, checks for the write permissions of the atKeys filepath. If the filePath has write permission, then continues with submission of enrollment request, else logs error and quits.

**- How I did it**
- Introduce "isWritable" method which attempts to create the directories for the given file if they do not exist, and then tries to open the file in write mode to verify write permissions. If the file can be opened for writing, it is immediately closed and deleted. In `AtOnboardingServiceImpl._generateAtKeysFile` method, there is a check which returns error if the file already exists. Therefore, delete the file here after checking for write permissions.
- Removed the logic that checks for the existence of the atKeys file from the enroll method, as this verification will now be handled by the isWritable method when creating the atKeys file to confirm write permissions.

**- How to verify it**
- Added the below unit tests to assert the behavior of the isWritable method:
   -  A test to verify isWritable returns false if directory has only read-only permissions
   - A test to verify isWritable returns false if file already exists in the directory
   - A test verify isWritable returns true if directory does not have a file already

- fix: Check of write permission of atKeys file path when submitting enroll request

### DIFF
--- a/packages/at_onboarding_cli/CHANGELOG.md
+++ b/packages/at_onboarding_cli/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 1.7.1
-- When submitting an enrollment request, check for write permissions of AtKeys file path.
+
+- fix: When submitting an enrollment request, check for write permissions of AtKeys file path.
 ## 1.7.0
 - feat: add `auto` command to the activate CLI 
 ## 1.6.4
--- a/packages/at_onboarding_cli/CHANGELOG.md
+++ b/packages/at_onboarding_cli/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 1.7.1
+- When submitting an enrollment request, check for write permissions of AtKeys file path.
 ## 1.7.0
 - feat: add `auto` command to the activate CLI 
 ## 1.6.4
--- a/packages/at_onboarding_cli/lib/src/cli/auth_cli.dart
+++ b/packages/at_onboarding_cli/lib/src/cli/auth_cli.dart
@@ -350,8 +350,16 @@ Future<void> enroll(ArgResults argResults, {AtOnboardingService? svc}) async {
   }
 
   File f = File(argResults[AuthCliArgs.argNameAtKeys]);
-  if (f.existsSync()) {
-    stderr.writeln('Error: atKeys file ${f.path} already exists');
+
+  // "_isWritable" attempts to create the directories for the given [file] if they do not exist,
+  // and then tries to open the file in write mode to verify write permissions.
+  //
+  // If the file can be opened for writing, it is immediately closed and deleted.
+  // In [AtOnboardingServiceImpl._generateAtKeysFile] method, there is a check which returns error
+  // if the file already exists. Therefore, delete the file here after checking for write permissions.
+  //
+  // Incase of any exceptions, the error is logged and returned.
+  if (isWritable(f) == false) {
     return;
   }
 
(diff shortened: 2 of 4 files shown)


gkc (changes requested):
@sitaram-kalluri I think the logic here is faulty. Consider the following scenario. 
. User onboards an atsign using onboard command - standard atKeys file created
. On the same host, the user attempts an enrollment using the enroll command leaving the keys file as default
. isWritable confirms the standard atKeys file is writable, and then deletes it
. the original atKeys file is now gone

sitaram-kalluri:
> @sitaram-kalluri I think the logic here is faulty. Consider the following scenario. . User onboards an atsign using onboard command - standard atKeys file created . On the same host, the user attempts an enrollment using the enroll command leaving the keys file as default . isWritable confirms the standard atKeys file is writable, and then deletes it . the original atKeys file is now gone

@gkc : The keys file is mandatory for the enroll command. Attaching the log snippet below
```
sitaram@sitaram-ThinkPad-E14:~/IdeaProjects/atsign/core/at_libraries/packages/at_onboarding_cli/bin$ dart activate_cli.dart enroll -s ABC123 -p wavi -d local -n wavi:rw -a @alice🛠 -r vip.ve.atsign.zone
Argument error for command enroll: The --keys option is mandatory for the "enroll" command
Usage: enroll
    -s, --passcode (mandatory)      The passcode to present with this enrollment request.
```

The below condition in the enroll method ensure the atKeys file path is mandatory:
```dart
@visibleForTesting
Future<void> enroll(ArgResults argResults, {AtOnboardingService? svc}) async {
  if (!argResults.wasParsed(AuthCliArgs.argNameAtKeys)) {
    throw ArgumentError('The --${AuthCliArgs.argNameAtKeys} option is'
        ' mandatory for the "enroll" command');
  }
  ```

Further, if the user gives the same atKeys filepath for enroll command as of the original atKeys, then it would return an error stating the atKeys file exists but does not delete the original atKeys file.

File: packages/at_onboarding_cli/CHANGELOG.md [hunk 1]
```diff
@@ -1,3 +1,5 @@
+## 1.7.1
+- When submitting an enrollment request, check for write permissions of AtKeys file path.
```
gkc: add 'fix: ' at the start of this comment please

gkc (changes requested):
@sitaram-kalluri you state
> Further, if the user gives the same atKeys filepath for enroll command as of the original atKeys, then it would return an error stating the atKeys file exists but does not delete the original atKeys file.

However, you have removed that specific check for the file already existing

packages/at_onboarding_cli/lib/src/cli/auth_cli.dart
```diff
@@ -350,8 +350,16 @@ Future<void> enroll(ArgResults argResults, {AtOnboardingService? svc}) async {
   }
 
   File f = File(argResults[AuthCliArgs.argNameAtKeys]);
-  if (f.existsSync()) {
-    stderr.writeln('Error: atKeys file ${f.path} already exists');
```
gkc, line 362: The above check should not have been removed

gkc: For maintainability purposes. It needs to be made clear to the reader precisely what is happening here. Burying this as an exception in a method called 'isWritable' makes the code confusing

sitaram-kalluri: Sure Gary, Added the "f.existsSync()" check back.

packages/at_onboarding_cli/CHANGELOG.md
sitaram-kalluri: Added "fix:" in the changelog.md

packages/at_onboarding_cli/lib/src/cli/auth_cli.dart
```diff
@@ -430,27 +435,21 @@ Future<void> enroll(ArgResults argResults, {AtOnboardingService? svc}) async {
...
 /// - Any other exceptions are caught and logged, indicating a failure to determine write access.
 @visibleForTesting
 bool isWritable(File file) {
   try {
     // If the directories do not exist, create them.
     // "recursive" is set to true to ensure that any missing parent directories are created.
-    // "exclusive" is set to true to check if the file already exists; if it does,
-    // an error will be logged and returned.
-    file.createSync(recursive: true, exclusive: true);
+    file.createSync(recursive: true);
```
gkc: I believe we should keep the `exclusive:true` here also, but allow the PathExistsException to be thrown. I'm a bit uncomfortable with a method called 'isWritable' possibly deleting a file

Maybe a better approach would be for this method to handle each case (file exists, file doesn't exist) differently
- if file exists, try opening it for FileMode.writeOnlyAppend
- if file doesn't already exist, try creating it and then removing it

gkc (changes requested):

packages/at_onboarding_cli/lib/src/cli/auth_cli.dart
sitaram-kalluri: @gkc : 
- Renamed "isWritable" method name to "canCreateFile".
- Added `exclusive:true` when creating a file and rethrowing the PathExistsException.

Please review and let me know if any further changes are required.

gkc (approved):

PR #711, opened 2024-11-18
fix: Changes in at_commons related to public key hash and implement sha512 hashing algo in at_chops

sitaram-kalluri:
**- What I did**
Changes in at_chops:

 - Implement SHA512 hashing algorithm in at_chops.
 - In at_chops_impl.dart, introduce "hashWith" static method which will take a hashing algorithm as input and return the hashed value the given input.

Changes in at_commons:
 - In PublicKeyHash.dart,  replace "PublicKeyHashingAlgo" enum with String which holds the hashing algo. This is because the "PublicKeyHashingAlgo" is a duplicate of "HashingAlgoType" enum and have an overhead of converting PublicKeyHashingAlgo type to HashingAlgoType in at_chops to hash the data. 
- Updated the existing unit tests accordingly.

### DIFF
--- a/packages/at_chops/lib/src/algorithm/default_hashing_algo.dart
+++ b/packages/at_chops/lib/src/algorithm/default_hashing_algo.dart
@@ -8,3 +8,11 @@ class DefaultHash implements AtHashingAlgorithm<List<int>, String> {
     return md5.convert(data).toString();
   }
 }
+
+class SHA512HashingAlgo implements AtHashingAlgorithm<List<int>, String> {
+  @override
+  String hash(List<int> data, {covariant HashParams? hashParams}) {
+    Digest digest = sha512.convert(data);
+    return digest.toString();
+  }
+}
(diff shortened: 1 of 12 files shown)


packages/at_chops/lib/src/algorithm/default_hashing_algo.dart
```diff
@@ -8,3 +8,11 @@ class DefaultHash implements AtHashingAlgorithm<List<int>, String> {
     return md5.convert(data).toString();
   }
 }
+
+class SHA512HashingAlgo implements AtHashingAlgorithm<List<int>, String> {
+  @override
+  String hash(List<int> data, {covariant HashParams? hashParams}) {
```
murali-shris, line 14: please write unit test for this hash method

murali-shris (changes requested):

packages/at_chops/lib/src/algorithm/default_hashing_algo.dart
sitaram-kalluri: Added unit test for SHA512 hashing algo.

murali-shris (approved):

PR #731, opened 2024-12-09
feat: new "at_policy" package for building policy services on atProtocol

gkc:
**- What I did**
- new package at_policy
- added at_policy to the build / test matrix in .github workflows
- added core unit tests

The at_policy library provides generic scaffolding for
- building policy management services
- with which policy enforcement endpoints communicate via atProtocol 
- using the atClient libraries
- and therefore get all the benefits of using atProtocol and the atClient libraries - outbound communication only, end-to-end encryption, using atSigns rather than IP addresses, etc

**- How I did it**
- Added core code
- Added tests
- Added main README
- Added examples and a README in the examples directory

**- How to verify it**
Tests pass

### DIFF
--- a/packages/at_policy/README.md
+++ b/packages/at_policy/README.md
@@ -1,0 +1,25 @@
+<a href="https://atsign.com#gh-light-mode-only"><img width=250px src="https://atsign.com/wp-content/uploads/2022/05/atsign-logo-horizontal-color2022.svg#gh-light-mode-only" alt="The Atsign Foundation"></a><a href="https://atsign.com#gh-dark-mode-only"><img width=250px src="https://atsign.com/wp-content/uploads/2023/08/atsign-logo-horizontal-reverse2022-Color.svg#gh-dark-mode-only" alt="The Atsign Foundation"></a>
+
+[![pub package](https://img.shields.io/pub/v/at_policy)](https://pub.dev/packages/at_policy)
+[![pub points](https://img.shields.io/pub/points/at_policy?logo=dart)](https://pub.dev/packages/at_onboarding_cli/score)
+[![gitHub license](https://img.shields.io/badge/license-BSD3-blue.svg)](./LICENSE)
+
+# at_policy
+
+## Introduction
+The at_policy libraries provide generic scaffolding for building policy 
+management services which policy enforcement endpoints communicate with via 
+atProtocol and therefore get all the benefits of using atProtocol - outbound 
+communication only, end-to-end encryption, using atSigns rather than IP 
+addresses
+
+## Usage
+See full worked examples in the [example](example/README.md) directory.
+
+## Open source usage and contributions
+
+This is freely licensed open source code, so feel free to use it as is, suggest
+changes or enhancements or create your own version.
+See [CONTRIBUTING.md](../../CONTRIBUTING.md) for
+detailed guidance on how to set up tools, tests and make a pull request.on. 
+See CONTRIBUTING.md for detailed guidance on how to set up tools, tests and make a pull request.
--- a/packages/at_policy/README.md
+++ b/packages/at_policy/README.md
@@ -7,7 +7,7 @@
 # at_policy
 
 ## Introduction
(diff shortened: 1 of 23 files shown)


cpswan (dismissed):
What's the plan to add this to the automated tests we run from [`at_libraries/.github/workflows/at_libraries.yaml`](https://github.com/atsign-foundation/at_libraries/blob/trunk/.github/workflows/at_libraries.yaml)?

gkc:
> What's the plan to add this to the automated tests we run from [`at_libraries/.github/workflows/at_libraries.yaml`](https://github.com/atsign-foundation/at_libraries/blob/trunk/.github/workflows/at_libraries.yaml)?

[Already done](https://github.com/atsign-foundation/at_libraries/pull/731/files#diff-7f2762ee0a2041b4e4ebd823359896fde1dbf7b1a6e13d8799ac81146e8ddb5d)

gkc:
@XavierChanth just noticed that I hadn't updated `example/bin/service.dart` to use PolicyService, which I added after the initial implementation in order to simplify the policy service development experience.

gkc:
@XavierChanth just noticed that I hadn't updated `example/bin/service.dart` to use PolicyService, which I added after the initial implementation in order to simplify the policy service development experience ... I'll update this tomorrow morning, maybe this evening although unlikely

gkc:
> @XavierChanth just noticed that I hadn't updated `example/bin/service.dart` to use PolicyService, which I added after the initial implementation in order to simplify the policy service development experience ... I'll update this tomorrow morning, maybe this evening although unlikely

Actually that's incorrect - `policy.dart` is using PolicyService - `service.dart` is not because it is the service which is calling the policy service. In a version 1.1 I may add a PolicyClient to make it easier to write clients of the policy service

packages/at_policy/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Usage
+See full worked examples in the [example](example/README.md) directory.
+
+## Open source usage and contributions
+
+This is freely licensed open source code, so feel free to use it as is, suggest
+changes or enhancements or create your own version.
+See [CONTRIBUTING.md](../../CONTRIBUTING.md) for
+detailed guidance on how to set up tools, tests and make a pull request.on.
```
xavierchanth: nit: typo

packages/at_policy/lib/src/policy/models.dart
```diff
@@ -0,0 +1,191 @@
...
+  Map<String, dynamic> toJson() => _$PolicyResponseToJson(this);
+
+  static PolicyResponse fromJson(Map<String, dynamic> json) =>
+      _$PolicyResponseFromJson(json);
+
+  @override
+  String toString() => jsonPrettyPrinter.convert(toJson());
+}
+
+abstract class CoreDeviceInfo {
```
xavierchanth: Maybe fromJson and toJson should be included as abstract members here?

packages/at_policy/pubspec.yaml
```diff
@@ -0,0 +1,31 @@
...
+  at_client: ^3.3.0
+  at_cli_commons: ^1.2.0
+  at_utils: ^3.0.19
+  json_annotation: ^4.9.0
+
+dependency_overrides:
+  dartssh2:
+    git:
+      url: https://github.com/atsign-foundation/dartssh2
+      ref: trunk
```
xavierchanth: This can be removed

xavierchanth (dismissed):

packages/at_policy/lib/src/policy/models.dart
gkc: Maybe in a future PR

packages/at_policy/README.md
gkc: fixed

packages/at_policy/pubspec.yaml
gkc: done

xavierchanth (approved):

PR #736, opened 2024-12-11
fix: remove 'transformers' from PolicyService interface and impl

gkc:
**- What I did**
fix: removed `requestTransformer` and `responseTransformer` from PolicyService and PolicyServiceImpl which were introduced in order to support legacy NoPorts authorization checks - this legacy support is now being handled within the NoPorts policy service

**- How I did it**
See commits

**- How to verify it**
Tests pass

### DIFF
--- a/packages/at_policy/lib/src/policy/impl.dart
+++ b/packages/at_policy/lib/src/policy/impl.dart
@@ -35,12 +35,6 @@ class PolicyServiceImpl with AtClientBindings implements PolicyService {
   @override
   final bool allowAll;
 
-  @override
-  RpcTransformer? requestTransformer;
-
-  @override
-  RpcTransformer? responseTransformer;
-
   late final AtRpc rpc;
 
   static const JsonEncoder jsonPrettyPrinter = JsonEncoder.withIndent('    ');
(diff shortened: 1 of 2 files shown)


packages/at_policy/lib/src/policy/impl.dart
```diff
@@ -130,32 +117,25 @@ class PolicyServiceImpl with AtClientBindings implements PolicyService {
...
       logger.severe('Exception: $e');
-      policyResponse = PolicyResponse(
+      rpcResponse = AtRpcResp(
+        reqId: rpcRequest.reqId,
+        respType: AtRpcRespType.error,
+        payload: PolicyResponse(
+          message: 'Exception: $e',
+          policyDetails: [],
+        ).toJson(),
         message: 'Exception: $e',
```
xavierchanth, line 137: Is message supposed to be there twice? I see it was moved into the payload, but also kept in response.

xavierchanth (approved):

packages/at_policy/lib/src/policy/impl.dart
gkc: Yes supposed to be there twice, thanks Xavier

PR #744, opened 2024-12-18
feat: at_chops faster aes impl

### LINKED ISSUE #705: feat (atchops): Faster aes ctr algorithms
Based on previous discussions and benchmarking, we plan to roll this out in two phases:

A note from me:

```
I believe the best path forward is:

Soon:
Decide on the algorithm (I think it's pretty clear that better is better for our needs)
Make non-breaking changes to uptake the new algorithm in pure-dart

When we have free time or more demand:
Make the breaking changes to atChops to return FutureOr<T> for all operations which 
could be asynchronous as a result of hardware acceleration
Propagate those changes to our packages
Add a new "initialize" function to at_client_mobile which is expected to be called 
at Flutter startup and loads the hardware acceleration libraries
```

Benchmarks were done against better_cryptography, webcrypto (boring ssl), cryptography, our C implementation of atchops using ffi (mbedtls) and the current library (encrypt)
The benchmarks were done without PKCS padding, since some algorithms don't include it out of the box.

Better cryptography was consistently the fastest of the stable algorithms, it doesn't support padding, so we will have to take that into consideration when implementing.

murali-shris:
Fixes https://github.com/atsign-foundation/at_libraries/issues/705
**- What I did**
- replaced current AES implementation with better_cryptography impl
**- How I did it**
- better_cryptography AES impl doesn't have padding logic implemented. Added abstractions for padding algo and implemented PKCS7Padding
- Modified returns types in AtEncryptionAlgorithm to FutureOr
- Added unit tests to check backward compatibility (data encrypted with old AES impl should be decrypted with better_crypto AES impl and vice versa
- Added unit tests
**- How to verify it**
- tests should pass
- at_client and server tests should pass with this branch of at_chops

### DIFF
--- a/packages/at_chops/lib/src/algorithm/padding/pkcs7padding.dart
+++ b/packages/at_chops/lib/src/algorithm/padding/pkcs7padding.dart
@@ -1,0 +1,51 @@
+import 'package:at_chops/src/algorithm/padding/padding.dart';
+import 'package:at_chops/src/algorithm/padding/padding_params.dart';
+import 'package:at_commons/at_commons.dart';
+
+class PKCS7Padding implements PaddingAlgorithm {
+  final PaddingParams _paddingParams;
+  PKCS7Padding(this._paddingParams);
+  @override
+  List<int> addPadding(List<int> data) {
+    if (_paddingParams.blockSize == null ||
+        _paddingParams.blockSize! <= 0 ||
+        _paddingParams.blockSize! > 255) {
+      throw AtEncryptionException('Block size must be between 1 and 255.');
+    }
+
+    // Calculate the number of padding bytes needed
+    int padding =
+        _paddingParams.blockSize! - (data.length % _paddingParams.blockSize!);
+
+    // Add padding bytes to the data
+    List<int> paddedData = List.from(data);
+    paddedData.addAll(List.filled(padding, padding));
+
+    return paddedData;
+  }
+
+  @override
+  List<int> removePadding(List<int> data) {
+    if (data.isEmpty) {
+      throw AtDecryptionException('Encrypted data cannot be empty');
+    }
+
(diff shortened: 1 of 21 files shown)


murali-shris:
Order of merge and publish
1. Merge at_chops PR and publish v3.0.0
2. Merge at_auth PR https://github.com/atsign-foundation/at_libraries/pull/745/files and publish at_auth v2.1.0
3. Merge at_client PR https://github.com/atsign-foundation/at_client_sdk/pull/1446 and publish at_client v3.5.0
4. Uptake at_chops, at_auth and at_client in at_onboarding_cli and publish

murali-shris:
@XavierChanth Can you please review the pkcs7 padding impl and the unit tests.Any additional test scenarios to be covered?

packages/at_chops/lib/src/algorithm/padding/pkcs7padding.dart
```diff
@@ -0,0 +1,49 @@
+import 'package:at_chops/src/algorithm/padding/padding.dart';
+import 'package:at_chops/src/algorithm/padding/padding_params.dart';
+import 'package:at_commons/at_commons.dart';
+
+class PKCS7Padding implements PaddingAlgorithm {
```
xavierchanth, line 5: LGTM, relevant RFC: [RFC 5652](https://datatracker.ietf.org/doc/html/rfc5652#section-6.3)

xavierchanth (approved):
LGTM, thanks!

xavierchanth:
> @XavierChanth Can you please review the pkcs7 padding impl and the unit tests.Any additional test scenarios to be covered?

It looks like you've already covered the most important edge cases, I think we have good enough coverage.

gkc:
@murali-shris is it OK to merge this PR to trunk?

PR #747, opened 2024-12-20
fix: at_lookup - flush socket after write and rethrow any exceptions …

### LINKED ISSUE #716: exception handling in at_lookup's BaseConnection.write
### Describe the bug

While looking at [this code](https://github.com/atsign-foundation/at_libraries/blob/trunk/packages/at_lookup/lib/src/connection/base_connection.dart#L54-L66) in the course of investigating another issue, I observed two problems
(1) exception is silently swallowed
(2) write() is not followed by flush()

### Steps to reproduce

n/a

### Expected behavior

- if we want to catch exceptions here rather than via the socket.done error handler, we should `await socket.flush()` after the call to `socket.write()`
- we should rethrow exceptions if they occur

Being part of the core client-side network comms code, changes here have a large blast radius if they introduce a new problem so while the impl here is straightforward, the testing will need to be comprehensive

### Screenshots

_No response_

### Smartphones

_No response_

_No response_

### Additional context

_No response_

sitaram-kalluri:
*- What I did*
- In at_lookup, connection.dart : 
    - call socket.flush() after socket.write()
    - Rethrow the exception from the write method.

Fixes #716

*- How to verify it*
- Created a draft [PR](https://github.com/atsign-foundation/at_client_sdk/pull/1457) in at_client pointing to this at_lookup commit and run all functional and E2E tests. All tests passed
- All the functional tests in the at_onboarding_cli passed.

*- Description for the changelog*
- fix: at_lookup - flush socket after write and rethrow any exceptions

### DIFF
--- a/packages/at_lookup/CHANGELOG.md
+++ b/packages/at_lookup/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 3.0.50
+- fix: Flush socket after write and rethrow any exceptions occurred 
 ## 3.0.49
 - build[deps]: Upgraded the following packages:
   - at_commons to v5.0.0
--- a/packages/at_lookup/lib/src/connection/base_connection.dart
+++ b/packages/at_lookup/lib/src/connection/base_connection.dart
@@ -59,9 +59,11 @@ abstract class BaseConnection extends AtConnection {
     }
     try {
       getSocket().write(data);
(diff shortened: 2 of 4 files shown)


gkc (approved):
LGTM, thank you

PR #769, opened 2025-02-18
feat(atcommons): Atsign string extensions

xavierchanth:
**- What I did**

- Added Atsign string extensions to replace AtUtils.fixAtsign.
  - Using an extension on String allows for values to be null coalesced which previous wasn't possible
- AtUtils.fixAtsign now calls String.toAtsign (moved to another PR so that at_commons can be published first)

**- How I did it**

**- How to verify it**

feat(atcommons): Atsign string extensions

### DIFF
--- a/packages/at_commons/lib/atsign.dart
+++ b/packages/at_commons/lib/atsign.dart
@@ -1,0 +1,73 @@
+import 'package:at_commons/at_commons.dart';
+
+/// A fully qualified atSign (e.g. @alice)
+typedef Atsign = String;
+
+/// An atSign without the "@" (e.g. alice)
+typedef AtsignWithoutAt = String;
+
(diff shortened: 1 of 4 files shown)


packages/at_commons/lib/atsign.dart
```diff
@@ -0,0 +1,73 @@
+import 'package:at_commons/at_commons.dart';
+
+/// A fully qualified atSign (e.g. @alice)
+typedef Atsign = String;
```
xavierchanth, line 4: Dart 2.13 feature, we are already `>=2.15`

gkc (approved):
LGTM. Will need another PR later to move a subset of at_utils unit tests (which are currently testing AtUtils.fixAtSign) to the at_commons unit test suite

PR #770, opened 2025-02-18
breaking!: Reorganize atchops

xavierchanth:
**- What I did**

Breaking:
- Removed AtEncryptionKeyPair and AtPkamKeyPair in favor of the abstract AtSymmetricKeyPair
  (although breaking, today it only requires an additional cast in some places from `AtSymmetricKeyPair` to `AtRSAKeyPair?`)
  However, this future proofs the API in preparation for supporting different key types / algos for pkam and shared encryption, it is actually the other parts of the API that need to be adapted which will element the need for the cast
- Corrected type naming inconsistencies (breaking)

Non-breaking:
- split algorithms directory into encryption,signing,hashing,padding
- moved all at_platform specific stuff into at_platform directory
- Modified keys stored in classes to decode to raw bytes on construction so we don't decode every time we do an operation
- Added constructor / getters for the keys so that they offer both raw bytes & base64 as options, preserving backward compatibility on the interfaces.

**- How I did it**

**- How to verify it**

breaking!: Reorganize atchops


packages/at_chops/example/at_chops_example.dart
```diff
@@ -36,7 +37,8 @@ void main(List<String> args) async {
     atChops = AtChopsImpl(atChopsKeys);
   }
 
-  var atEncryptionKeyPair = atChops.atChopsKeys.atEncryptionKeyPair;
+  var atEncryptionKeyPair =
+      atChops.atChopsKeys.atEncryptionKeyPair as AtRSAKeyPair?;
```
xavierchanth, line 41: Example of the place we need the cast, atChopsKeys has been generalized

packages/at_chops/lib/at_chops.dart
```diff
@@ -1,34 +1,16 @@
-library at_chops;
+// export cryptography wrappers
+export 'src/hashing/hashing.dart';
```
xavierchanth, line 2: Added files named after the directories in src to make export management easier.

packages/at_chops/lib/src/at_keys_crypto.dart
```diff
@@ -65,7 +66,7 @@ class _AtKeysCryptoImpl implements AtKeysCrypto {
     String hashKey =
         await _getHashKey(passPhrase, _hashingAlgoType, hashParams: hashParams);
 
-    AESKey aesKey = AESKey(hashKey);
+    AtAESKey aesKey = AtAESKey(hashKey);
```
xavierchanth, line 69: renamed to prevent namespace collisions with other packages.

packages/at_chops/lib/src/at_platform/at_chops.dart
```diff
@@ -2,18 +2,18 @@ import 'dart:async';
...
+import 'package:at_chops/src/model/signing_result.dart';
 
 /// Base class for all Cryptographic and Hashing Operations. Callers have to either implement
 /// specific encryption, signing or hashing algorithms. Otherwise default implementation of specific algorithms will be used.
 abstract class AtChops {
-  final AtChopsKeys _atChopsKeys;
+  AtChopsKeys get atChopsKeys;
 
-  AtChopsKeys get atChopsKeys => _atChopsKeys;
+  const AtChops.init();
```
xavierchanth, line 12: This constructor does nothing, it only exists to allow the class to be extended.

packages/at_chops/lib/src/at_platform/at_chops_keys.dart
```diff
@@ -1,27 +1,27 @@
...
+  AsymmetricKeyPair? atSigningKeyPair;
 
   /// Default self encryption key
   SymmetricKey? selfEncryptionKey;
 
   /// APKAM symmetric key created during new enrollment
   SymmetricKey? apkamSymmetricKey;
 
+  /// EnrollmentId associated with pkam keys
+  String? enrollmentId;
```
xavierchanth, line 20: Added enrollmentId so that AtAuthKeys in at_auth can be casted to and from AtChopsKeys (without losing information).

packages/at_chops/lib/src/hashing/argon2id_hashing_algo.dart
```diff
@@ -50,3 +49,37 @@ class Argon2idHashingAlgo implements AtHashingAlgorithm<String, String> {
...
+/// A class that holds the parameters for configuring a hashing algorithm.
+///
+/// This class is used to customize the behavior of a hashing algorithm by
+/// providing control over key parameters such as parallelism, memory usage,
+/// iteration count, and the length of the resulting hash.
+///
+/// These parameters are particularly useful when working with algorithms
+/// like Argon2id, which can be adjusted for performance and security needs.
+
+class ArgonHashParams extends HashParams {
```
xavierchanth, line 62: Moved from a separate file to be with it's hashing algorithm implementation

packages/at_chops/lib/src/hashing/at_hashing_algorithm.dart
```diff
@@ -0,0 +1,27 @@
...
+
+/// Interface for hashing data. Refer [DefaultHash] for sample implementation.
+abstract class AtHashingAlgorithm<K, V> {
+  /// Hashes the passed data
+  FutureOr<V> hash(K data, {covariant HashParams? hashParams});
+}
+
+class HashParams {}
+
+class DefaultHashingAlgo extends Md5HashingAlgo {}
```
xavierchanth, line 14: I've taken all the DefaultXXXAlgo and named them after their proper algorithm, then extended each of them to become the new DefaultAlgo. They still all conform to the same interface.

packages/at_chops/lib/src/key/asymmetric_key_pair.dart
```diff
@@ -0,0 +1,31 @@
+import 'dart:typed_data';
+
+/// Represents a key pair for asymmetric public-private key encryption/decryption
+abstract class AsymmetricKeyPair<Pub extends AtPublicKey,
```
xavierchanth, line 4: AsymmetricKeyPairs can now be flavored with covariants of AtPublicKey and AtPrivateKey. Non breaking change, as they still conform to the same interface.

packages/at_chops/lib/src/key/at_aes_key.dart
```diff
@@ -1,31 +1,37 @@
...
+import 'package:at_chops/src/key/symmetric_key.dart';
 import 'package:encrypt/encrypt.dart';
 import 'dart:convert';
 
 /// Represents an AES key for symmetric encryption.
-class AESKey extends SymmetricKey {
-  final String _aesKey;
+class AtAESKey extends SymmetricKey {
+  @override
+  final Uint8List raw;
```
xavierchanth, line 10: Changed so that the stored member is the raw bytes rather than the base64 encoded string. This should reduce decrypt/encrypt latency as we don't need to decode every time now.

packages/at_chops/lib/src/key/key_type.dart
```diff
@@ -1 +1,11 @@
-enum EncryptionKeyType { rsa2048, rsa4096, ecc, aes128, aes192, aes256 }
+enum EncryptionKeyType {
```
xavierchanth, line 1: This is going to be a blocker later...

Sometimes we want to infer an algo, but when we infer from a runtimeType we don't always have the key size, nor care about the key size.

But also, when we get the algo used to encrypt / sign from a string, it's useful to have the algo. It's questionable whether it's useful to have the size. Maybe we need to make this a sealed class later, but, then it's not open to extension...

xavierchanth:
Setting back to draft, more work to be done.

PR #781, opened 2025-04-02
fix: (at_onboarding_cli): write keys before deleting cram secret 

### LINKED ISSUE #763: fix (at_onboarding_cli): write keys before deleting cram secret
Reason: Keys must be written before deleting the cram secret, otherwise the atServer must be reset to continue.

The decision here has an impact on a change to atServer that @gkc will be making:
If there is an approved firstApp enrollment, and the cram secret is still in place, then new firstApp enrollments will replace the existing one. (Currently they are denied)

- We can either prefer immediately before deleting cram secret
  - I think this is preferred once the above atServer change is made, as if there is a failure, in one of the operations, the keys will not be overridden and the re-running the cli will be able to self-recover
- Or we can do it immediately before deleting the cram secret
  - Preferred without the atServer change, because it can be manually recovered, but not automatically recovered by re-running the cli.

gkc:
**- What I did**
Fixes #763 
- at_auth:
  - feat: enable callers of `AtAuth.onboard` to control post-auth activation completion (set the encryption public key on the server, delete the "cram" secret)
- at_onboarding_cli:
  - fix: have `AtOnboardingServiceImpl.onboard` only perform post-auth activation completion once the atKeys file has been successfully saved
  - feat: Enable callers of `AtOnboardingServiceImpl.onboard` to control post-auth activation completion (set the encryption public key on the server, delete the "cram" secret)
- at_onboarding_cli_functional_tests:
  - test: Have the onboarding test call onboard three times instead of once:
    1. onboard and write atKeys file, but do not "complete activation" post-auth
    2. onboard again, and complete the activation - should succeed because activation was not completed before
    3. onboard again - will fail because the activation was previously completed

**- How I did it**
See commits

**- How to verify it**
Tests pass

### DIFF
--- a/packages/at_onboarding_cli/pubspec.yaml
+++ b/packages/at_onboarding_cli/pubspec.yaml
@@ -1,6 +1,6 @@
 name: at_onboarding_cli
 description: Dart tools for initial client onboarding, subsequent client enrollment, and enrollment management.
-version: 1.8.3
+version: 1.9.0
 repository: https://github.com/atsign-foundation/at_libraries
 homepage: https://atsign.com
 documentation: https://docs.atsign.com/
@@ -12,6 +12,9 @@ executables:
   at_register: register_cli
   at_activate: activate_cli
 
+dependency_overrides:
+  at_auth:
+    path: ../at_auth
 dependencies:
    args: ^2.5.0
    crypton: ^2.2.1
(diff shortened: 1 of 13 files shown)


packages/at_onboarding_cli/pubspec.yaml
```diff
@@ -12,6 +12,9 @@ executables:
   at_register: register_cli
   at_activate: activate_cli
 
+dependency_overrides:
+  at_auth:
+    path: ../at_auth
```
gkc, line 17: Will publish at_auth once this PR has been approved and merged. Will then remove this override in another PR before going on to publish at_onboarding_cli

xavierchanth (approved):

PR #790, opened 2025-04-30
feat: at_cli_commons: improve usability of CLIBase args management

gkc:
**- What I did**
feat: Improve usability for args management, including the ability to hide 
  some args which, while they are very helpful for dev purposes, are not so 
  friendly for end-users.

**- How I did it**
See diffs

### DIFF
--- a/packages/at_cli_commons/lib/src/cli_base.dart
+++ b/packages/at_cli_commons/lib/src/cli_base.dart
@@ -153,7 +153,7 @@ class CLIBase {
         verbose: parsedArgs['verbose'],
         syncDisabled: parsedArgs['never-sync'],
         maxConnectAttempts: int.parse(parsedArgs['max-connect-attempts']),
-        passPhrase: parsedArgs['passPhrase']);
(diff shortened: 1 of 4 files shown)


packages/at_cli_commons/lib/src/cli_base.dart
```diff
@@ -13,55 +13,129 @@ import 'package:version/version.dart';
...
+          negatable: false,
+          help: 'Do not run sync',
+          hide: hide.contains('never-sync'))
+      ..addOption('max-connect-attempts',
+          help: 'Number of times to attempt to initially connect to atServer.'
+              ' Note: there is a 3-second delay between connection attempts.',
+          defaultsTo: defaultMaxConnectAttempts.toString(),
+          hide: hide.contains('max-connect-attempts'))
+      ..addOption('pass-phrase',
+          aliases: ['passPhrase'],
```
gkc, line 97: Changed to 'pass-phrase' to match convention; aliased to 'passPhrase' for backwards compatibility

cpswan (approved):

PR #800, opened 2025-06-13
build: Use latest at_client in order to remove unnecessary direct dependency from at_onboarding_cli

gkc:
**- What I did**
build: Use latest at_client in order to remove unnecessary direct dependency from at_onboarding_cli

**- How I did it**
- eliminated use of `AtData` (which is part of the at_persistence packages) in at_onboarding_cli. AtData was being used because AtClient hadn't fully / correctly implemented the apkam authorization check such that it doesn't check for `local:` keys.

**- How to verify it**
Tests pass

### DIFF
--- a/tests/at_onboarding_cli_functional_tests/test/enrollment_test.dart
+++ b/tests/at_onboarding_cli_functional_tests/test/enrollment_test.dart
@@ -24,19 +24,32 @@ void main() {
   AtSignLogger.root_level = 'WARNING';
   String storageDir = 'test/storage';
 
+  String atSign1 = '@naresh🛠';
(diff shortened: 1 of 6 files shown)


tests/at_onboarding_cli_functional_tests/test/enrollment_test.dart
```diff
@@ -24,19 +24,32 @@ void main() {
   AtSignLogger.root_level = 'WARNING';
   String storageDir = 'test/storage';
```
gkc, line 25: No material changes in this file, just pulled the atSign variables out into main scope so it's easier to change them

cconstab (approved):

PR #806, opened 2025-07-03
feat: at_onboarding_cli: reuse the authenticated connection from AtAuth.authenticate

gkc:
**- What I did**
- take up at_client 3.5.0
- reuse the authenticated AtLookUp connection from AtAuth.authenticate, injecting it into the AtClient which is handed back to the calling code

**- How I did it**

**- How to verify it**
Tests pass

### DIFF
--- a/packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
+++ b/packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
@@ -4,7 +4,6 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
-import 'package:at_auth/at_auth.dart' as at_auth;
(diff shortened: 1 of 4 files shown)


packages/at_onboarding_cli/lib/src/onboard/at_onboarding_service_impl.dart
```diff
@@ -64,8 +63,10 @@ class AtOnboardingServiceImpl implements AtOnboardingService {
     await atClientManager.setCurrentAtSign(
         _atSign, atOnboardingPreference.namespace, atOnboardingPreference,
         atChops: atChops,
+        atLookUp: atLookUp,
```
gkc, line 66: This is the only actual change in this PR

cconstab (approved):

